PR #4, opened 2021-12-01
✨ show next open time when mint is closed

percentcer:
When the mint is closed, display when it will be next open (in the top bar)

image

### DIFF
--- a/src/components/Utils/ContractsOpened.tsx
+++ b/src/components/Utils/ContractsOpened.tsx
@@ -1,15 +1,36 @@
 import style from "./ContractsOpened.module.scss"
 import cs from "classnames"
 import { useContractStorage } from "../../utils/hookts"
-import { getNextOpenTime } from "../../utils/schedule"
+import { getMsUntilClose, getNextCycleStartTime } from "../../utils/schedule"
+import { useEffect, useState } from "react"
+import { getLocalTimezone } from "../../utils/timzones"
 
 export function ContractsOpened() {
-  const { data: issuerStorage } = useContractStorage(process.env.NEXT_PUBLIC_TZ_CT_ADDRESS_ISSUER!)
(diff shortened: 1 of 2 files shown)


ciphrd (changes requested):
It's a good feature, however this implementation does not provide the correct hour.

The environment variable OPENING_SCHEDULE_REFERENCE actually refers to a time of reference set in the past, from which are derived the cycle openings of the platform (the isPlatformOpened function is a good example).

So starting from OPENING_SCHEDULE_REFERENCE, there are cycles of 12 hours opening / 11 hours closing looping endlessly.

You would need to find the next time when the platform will open, based on the current time.

I would also suggest adding the timezone code along the date, as some people may have a local timezone not properly set to the timezone of their geographic position.

percentcer:
Ah thanks for the info, let me take another stab at it! Do you mind posting your current `OPENING_SCHEDULE_REFERENCE` as well? I wasn't sure if I had set a reasonable value in my `.env.local`

percentcer:
Updated commit to use proper cycle time calculations. Also added the timezone abbreviation and included the closing time information (when mint is already open)

image

ciphrd:
Thanks for your work ! 
I ended up removing the timezone code because it was showing RDT for me (which is one hour off).
This is not related to your work but rather the timezone sources i'm using (the utc name is correct but the abbreviation isn't, which is really weird).

For reference this is the timezone in question:

```json
{
  "value": "Romance Standard Time",
  "abbr": "RDT",
  "offset": 2,
  "isdst": true,
  "text": "(UTC+01:00) Brussels, Copenhagen, Madrid, Paris",
  "utc": [
    "Africa/Ceuta",
    "Europe/Brussels",
    "Europe/Copenhagen",
    "Europe/Madrid",
    "Europe/Paris"
  ]
},
```

RDT is hour off from UTC+1, which doesn't seem right... Needs more investigation, this whole timezone thing is had to track

PR #5, opened 2021-12-01
chore: leverage useIsMounted hook

huang47:
Objective:
drop `ClientOnly` / `ClientOnlyEmpty`'s own implementation of isMounted check in favor of `useIsMounted` hook

Thanks for the clean and very readable code base and I do learn from it a lot :) 

Notes:
* I don't seem to find any tests being setup so it's probably too early to do this, please let me know if adding few tests would be appropriate atm
* It's uncertain to me if `hookts` is named this way intentionally so I just keep it


ciphrd:
Thank you for your contribution, merging.

PR #7, opened 2021-12-04
Update preview iframe allow permissions

liamegan:
Adding allow property to iframes to allow for access to various browser APIs from cross-domain iframes.

Allows for access to:
- accelerometer
- camera
- gyroscope
- microphone
- xr-spatial-tracking;


ciphrd:
Thanks for your contribution Liam !

PR #8, opened 2021-12-04
✨ optimize community reports page

huang47:
community reports page always hit the backend to retrieve content. Currently it's a public page and all members / visotrs might see the same content regardless. We can build the page and always serve the cache when appropriate and rebuild page periodically.


ciphrd:
Thanks you for your contribution, this is a well-thought optimization

PR #10, opened 2021-12-16
🐛 don't show calculated open/close time if the contract has been manually paused

percentcer:
This omits printing the open/close time string if the contract's `paused` field doesn't match the calculated value. Avoids confusing situations like this one:
image

instead just going back to the old style:
image


liamegan:
Closing as irrelevant.

PR #11, opened 2021-12-19
Added price to generative card so it displays on Explore Page.

adaptajay:
Adding price to explore page, so that users can easilly decide which pieces stand out based on interest and price. Without having to open every piece to find out the price.


ciphrd:
Regarding adding the price on the cards, I would prefer if it was an option.
I am currently implementing the dark mode, which also comes with settings stored in the local storage. I'll check this and eventually merge it so that it can be toggled by a setting too :)

PR #34, opened 2022-01-11
Ensure all content about the artwork wraps.

tensafefrogs:
This prevents content from overflowing off the screen and also fixes a bug on mobile where the window zoom is too small due to the overflowed content.


ciphrd:
I missed this PR - it's been adjusted when writing the v1.0
Thx for the contribution but no longer relevant (my bad sorry)

PR #53, opened 2022-02-16
Adding updated CoC to include section about PNG artworks

liamegan:


ciphrd:
I would add a small message saying that these are temporary measure until we implement a proper tagging system on the projects.

liamegan:
> I would add a small message saying that these are temporary measure until we implement a proper tagging system on the projects.

I don't think it IS a temporary measure, though. If a user, in future, makes an artwork that looks generative but is, in fact, a layered PNG then it's still in violation of this code. Yes we will have the ability to change the tag, but this is something that should ultimately be on the user to understand.

A larger part of the point of this particular code, in my mind, is the desire for artists to better describe their work.

Let me know if you disagree and I'll add the line, otherwise more than happy to talk some more about this.

ciphrd:
I was talking about the need to add it in the description itself.
However we can maybe push it like so and update it later on when we see fit

PR #57, opened 2022-04-20
Bugfix for error while scrolling a token's activity list

CiaccoDavide:
I'm referring to this error:

https://user-images.githubusercontent.com/3226096/164323669-1ece5d0f-a403-460a-9e5c-75769920d834.mp4

steps to reproduce:
 - open project page
 - click on the "activity" tab
 - scroll until error

it happens because actions of type `UPDATE_PRICING` and `UPDATE_STATE` have a `null` issuer, and the component `UserBadge` doesn't check if the user is `null`

(it happens only if in the fetched actions there are some of type `UPDATE_PRICING` and `UPDATE_STATE`, that's why on some pages it doesn't happen or it crashes only after older data is fetched while scrolling)

My solution: don't show UserBadge if user is undefined

I know it is not really a front-end issue because the issuer should (probably) always be returned, but this way it can't crash if it doesn't (like it does now)

here is how it looks "fixed":

20220420_160208_chrome_vtR0unG4Ol

_notice how "updated state" and "updated price" actions don't have a UserBadge on the left_

### DIFF
--- a/src/components/User/UserBadge.tsx
+++ b/src/components/User/UserBadge.tsx
@@ -18,8 +18,8 @@ interface WrapperProps {
   children: ReactNode
 }
 
-const WrapperLink = ({ 
-  className, 
+const WrapperLink = ({
+  className,
   user,
   children,
 }: WrapperProps) => (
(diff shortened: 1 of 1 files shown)


ciphrd:
LGTM, thanks 👍

PR #59, opened 2022-04-22
change/MasonryCards

jbmoutout:
@ciphrd A very basic approach to Masonry (avoiding the mandatory `height` Flexbox caveat).
Would require an eventListener to update number of columns on window resize.

Screenshot 2022-04-22 at 16 38 27

### DIFF
--- a/src/components/Card/CardsContainer.module.scss
+++ b/src/components/Card/CardsContainer.module.scss
@@ -5,8 +5,13 @@
   display: grid; 
   grid-template-columns: repeat(auto-fit, minmax(270px, 1fr));
   gap: var(--cards-gap);
+  grid-auto-flow: column;
 
   & > a {
     text-decoration: none !important;
+  }
+  .col {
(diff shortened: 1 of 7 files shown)


ciphrd:
Runtime error when running on the home page:

image


On the explore page (and all the other pages using the `CardsContainer` component), the cards are visually broken:

image


The implementation in itself is okay, but creating a new `MasonryCardsContainer` would have been more appropriate.

jbmoutout:
Agreed,was only attempting to show a quick lead for futur implementation

jbmoutout:
@ciphrd a bit more complete implementation example with dedicated component and resizing listener.

⚠️  issue between the CardLoading logic and masonry when replacing placeholder height (see loom) that would require more work. https://www.loom.com/share/19388a89e654497187473553e04a822b

maerzhase:
we can close this one, i opened anohter PR based on it #153

PR #77, opened 2022-04-25
Bug/#75 fix clickable areas

### LINKED ISSUE #75: Project page, clickable areas are bigger than visual area
The artist name, as well as some clickable areas in the project details are bigger than their visual size.
They take the full width while they should be the same size as what we see.

I suspect this is due to flex container stretching children.

maerzhase:
This PR fixes #75. Anchors previously received the container styling which causes the interactive area of the anchor to expand with e.g. `flexbox`, `grid`. Moving the container e.g. inside the anchor fixes the problem inside the `<UserBadge>` and simply wrapping them in a div around them in the `grid` styled components too. 

- fix #75 

> still in draft, another issue was reported and I wanna go trough the pages again to double-check

### DIFF
--- a/src/components/List/ListSplits.tsx
+++ b/src/components/List/ListSplits.tsx
@@ -32,8 +32,9 @@ export function ListSplits({
             <span className={cs(style.pct)}>
               {displayPercentage(item.pct/1000)}%
             </span>
-            <div className={cs(style.user)}>
-              <UserBadge
+            <div>
+	      <UserBadge
+		className={style.user}
                 size="small"
                 user={item.user}
                 displayAvatar={false}
--- a/src/components/User/UserBadge.tsx
+++ b/src/components/User/UserBadge.tsx
@@ -24,8 +24,10 @@ const WrapperLink = ({
   children,
 }: WrapperProps) => (
   <Link href={getUserProfileLink(user)}>
-    <a className={className}>
-      {children}
+    <a className={cs(style.link, className)}>
+      <div className={style.container}>
+	{children}
+      </div>
     </a>
   </Link>
 )
@@ -60,7 +62,7 @@ export function UserBadge({
 
   return user ? (
     <Wrapper
-      className={cs(style.container, style[`side-${avatarSide}`], className)}
+      className={cs({[style.container]: !hasLink}, style[`side-${avatarSide}`], className)}
       user={userAlias}
     >
       {displayAvatar && (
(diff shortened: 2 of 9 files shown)


ciphrd (changes requested):
I think wrapping the `UserBadge` is the right approach since we see the issue with the badge at multiple places throughout the app.
However I noticed that by adding a div it break some css rules used in particular components where the css properties of the badge are over written:

image

expected output:

image

We need an update on the `components/List/ListSplits` component to reflect the change properly.

I also noticed that it doesn't seem to fix the user badge on the collaboration dropdown:

image

Adding a `align-items: flex-start;` on the container of the dropdown in the `EntityBadge` component will most likely fix this issue.

ciphrd (changes requested):
I think a simple css rule on the parent container can do the job, and doesn't require adding a DOM element:

```css
.extra_details {
  display: grid;
  grid-template-columns: max-content 1fr;
  align-items: baseline;
  gap: 12px 5px;
  /* prevents children elements from streching */
  justify-items: start;
}
```

maerzhase:
@ciphrd thanks for your input. I reverted the commits that were adding the unnecessary DOM elements from the details. The solution with `justify-items: start` works! 👍

src/components/User/UserBadge.tsx
```diff
@@ -60,7 +62,11 @@ export function UserBadge({
 
   return user ? (
     <Wrapper
-      className={cs(style.container, style[`side-${avatarSide}`], className)}
+      className={cs(
+	{[style.container]: !hasLink, [style.default_font_styles]: !hasLink},
```
maerzhase, line 66: @ciphrd to make overwrites work consistently I couldn't find a smarter solution then extracting them in their own class and add them on the container or the anchor conditionally. 

I believe that this solution has the least dom polution and works consistently through the places we use the `<UserBadge />`

src/components/List/ListSplits.tsx
```diff
@@ -32,8 +32,9 @@ export function ListSplits({
             <span className={cs(style.pct)}>
               {displayPercentage(item.pct/1000)}%
             </span>
-            <div className={cs(style.user)}>
-              <UserBadge
+            <div>
+	      <UserBadge
+		className={style.user}
```
maerzhase, line 37: Since the anchor remains the outermost container in the `<UserBadge />` we can simply overwrite styles by applying the class to the component directly

ciphrd:
LGTM 👍

Issue #63, opened 2022-04-23
Scoped setting on the card containers to define the size of the cards in the view


ciphrd:
Users should be able to define the size of the cards when exploring cards. The setting should be scoped to each page, and persisted on the client-side.

![image](https://user-images.githubusercontent.com/18054507/164909351-81a639dd-c741-4006-9dac-07dd3817c782.png)

It should leverage the existing SettingsContext

The following icon can be used: https://fontawesome.com/icons/square?s=solid

maerzhase:
on it!

PR #90, opened 2022-04-27
Feature: scoped card container size

maerzhase:
@ciphrd here is the draft for the size select feature (#63). 

- #63

Thats how it looks right now:

Kapture 2022-04-27 at 19 24 49

There are a few things we should decide on before I proceed :-)
First of all did you have any sizes in mind? These are the values for the select Component I choose for now.

```ts
const cardSizeOptions: IOptions[] = [
  {
    label: 'small',
    value: 120,
  },
  {
    label: 'medium',
    value: 270,
  },
  {
    label: 'large',
    value: 400,
  },
]
```
As I implemented it at the moment there is only a size select component on the `/generative/` route. Therefore also only this `CardsExplorer` has a `cardSizeScope`. I am not sure which other views should get the option. e.g. marketplace probably just need the bigger cards, because of the information it should display?! Please also note that I am currently only storing one value in the `SettingsContext` and if no scope is provided we just fallback to the default (270). I would implement a map with scoped card sizes in the settings once we know which other views should get a card-size select. 

In general I still need to revise the style of the cards to work with smaller sizes but i would do that once we decided on the last formalities of the feature.


src/containers/Generative/Iterations/GenerativeIterations.module.scss
```diff
@@ -0,0 +1,38 @@
...
+  button {
+    width: 50px;
+    height: 50px;
+    margin: 0;
+    padding: 0;
+    border: 3px solid var(--color-border-input);
+    background: transparent;
+    color: var(--color-border-input);
+    cursor: pointer; 
+    &.active {
```
ciphrd, line 11: I would remove the pointer events on the active size button

ciphrd:
This implementation looks good to me, good job.
I would increase the value for the smallest cards as we can't see much and may need some more space for displaying informations (from my tests a value of 200 seems enough).

On a side note, i really need to add some prettifier rules to the project ahah

image

maerzhase:
@ciphrd i increased the size for small. and i also created a reusable component for the `CardSizeSelect`

maerzhase:
> I noticed a bug when loading a project page and then going back to the explore page: I think this is due to the /explore not having a card size scoped:

Actually its due to the user page not using the `CardsExplorer`. I've implemented a reset to default cardsize on effect cleanup. I think thats the simplest solution to ensure default card-size for other views too. 

> Also the pointer events should be disabled on the button of the active card size to prevent the hover effect.

✅

ciphrd (approved):
LGTM
I made a few stylistic changes

PR #91, opened 2022-04-27
add graphql for user entire collection

### LINKED ISSUE #86: Make the user collection enjoy fetch the whole collection
The current enjoy feature of the user collection is fetching the tokens with some pagination, so we can't randomize the order at first.

Instead of using `generativeToken { objkts(skip, take) {} }`, it should use `generativeToken { entireCollection }` instead.

This will require some changes to the component `UserCollectionEnjoy`, feeding the tokens to the `GenerativeEnjoy` component. 

Link to a page:
https://www.fxhash.xyz/u/ciphrd/collection/enjoy

maerzhase:
@ciphrd I implemented a separate `Qu_userEntireCollection` for the enjoy collection because `Qu_userObjkts` i also used in the normal user collection page and i guess there it makes sense to keep paginated objkts. let me know what you think

- fix #86

### DIFF
--- a/src/containers/Generative/Enjoy/GenerativeEnjoy.tsx
+++ b/src/containers/Generative/Enjoy/GenerativeEnjoy.tsx
@@ -1,18 +1,16 @@
 import style from "./GenerativeEnjoy.module.scss"
 import Link from "next/link"
 import cs from "classnames"
-import { GenerativeTokenWithCollection } from "../../../types/entities/GenerativeToken"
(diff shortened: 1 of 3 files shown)


ciphrd:
Update to the query looks good.
I added an optimization to:
- prevent refetch which was originally designed for the pagination
- inject the owner which was missing

And a loader for the initial fetch since it can take a while

ciphrd:
The metadata field in the objkt is not required anymore to display the live view, and so it can be removed from the query (reducing the size of the response payload by 2/3)

PR #98, opened 2022-04-28
feat(cards-container): display filters button only when visible in viewport

### LINKED ISSUE #83: [Mobile] The filter button should only be displayed the Cards Container is visible
We have a `Filters` button which opens the filter panel, displayed at the bottom of the screen on mobile:

image

This button should be hidden unless we the Cards Container is in the view. So whenever the red area in the following video is in the view:

https://user-images.githubusercontent.com/18054507/165501909-18296d26-9569-4411-81ba-7902803950a5.mp4

GemN:
fixes #83


ciphrd (approved):
LGTM

PR #99, opened 2022-04-28
feat(search): search can be minimized on mobile

### LINKED ISSUE #82: The search header takes too much space on mobile
The current header takes too much vertical space.

We would like the search bar and the sort dropdown to be on the same line. The search bar could be replaced by a button which would expand a search bar on the full width when clicked on.

GemN:
fixes #82 

A few notes:
I introduce two new hooks with this PR.
- useClickOutside: for component behavior purpose, allow to check if the user clicked outside an element
- useWindowsSize: for component behavior purpose, allow to check the window size in real time. 
Even if it's preferable to use css to handle breakpoints, sometimes we can't use css only to handle behavior (see component SearchInput)

### DIFF
--- a/src/components/Input/SearchInput.tsx
+++ b/src/components/Input/SearchInput.tsx
@@ -32,9 +32,7 @@ export function SearchInput({
   const isMobile = useMemo(() => width !== undefined && (width <= breakpoints.sm), [width]);
   const handleMinimize = useCallback((newState) => {
     setIsMinimizedOnMobile(newState);
-    if (onMinimize) {
-      onMinimize(newState);
-    }
+    onMinimize?.(newState);
   }, [onMinimize]);
   const handleSubmit = useCallback((event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
--- a/src/components/Input/SearchInput.module.scss
+++ b/src/components/Input/SearchInput.module.scss
@@ -7,7 +7,7 @@
   border: 3px solid var(--color-border-input);
   background-color: var(--color-white);
   height: 50px;
-  padding-left: 4px;
+  padding: 0 10px 0 4px;
 
   button {
     border: none;
@@ -15,6 +15,7 @@
   }
 
   input {
+    background: transparent;
     border: 0;
     font-family: inherit;
     outline: none;


GemN:
I thought:
- User writes his search, then click on sort by. However I just notice this process does not work currently because it would require to trigger search on blur. 

Your suggestion:
- Hide the sort when the input search is focused
The implementation should not be that hard because I can add a callback "onMinimize" that will send his state back to the parent.

Should I minimise the sort or completely hide it ? 
Completely hide it force this user flow:
- user write his search
- user "enter"
- page scroll to top marker, the search unfocus
- user can select a sort

Minimizing the sort allow this user flow:
- user write his search
- user click on the sort minimized
- the search trigger on blur, the input minimize, the sort expand and user can select his sort choice
(it saves a click)

At the moment I'd rather do the first userflow because i'm not sure we want to trigger the search on blur?

ciphrd:
I think we really don't want to trigger a search on blur (not an expected UX pattern).

While it would make sense to trigger a search on blur **if and only if** the minimized sort is clicked after, I think it makes the UX weird and would require a few implementation tricks (to prevent the auto scroll for instance).

I'd rather opt for hiding the sort altogether when the search input is focused, forcing the trigger of the search with "enter". And then if the user wants a different sort they can change it.

src/components/Input/SearchInput.tsx
```diff
@@ -21,29 +22,36 @@ export function SearchInput({
...
 }: Props) {
   const refInput = useRef<HTMLInputElement>(null);
   const refForm = useRef<HTMLFormElement>(null);
   const [isMinimizedOnMobile, setIsMinimizedOnMobile] = useState(minimizeOnMobile);
   const { width } = useWindowSize();
   const isMobile = useMemo(() => width !== undefined && (width <= breakpoints.sm), [width]);
+  const handleMinimize = useCallback((newState) => {
+    setIsMinimizedOnMobile(newState);
+    if (onMinimize) {
+      onMinimize(newState);
```
ciphrd, line 36: Can be onlined 
```js
onMinimize?.(newState)
```

ciphrd (changes requested):
Small detail: the change breaks the color of the search input background on dark theme:
image

ciphrd (approved):
LGTM

Issue #66, opened 2022-04-23
add warning to Profile update experience that tzprofiles updates may take a few days to populate on site


ciphrd:

Cerfio:
On it !

liamegan:
Unassigning

PR #100, opened 2022-04-29
add warning after a profile update

Cerfio:
fix https://github.com/fxhash/fxhash-website/issues/66


ciphrd (changes requested):
Oh, my bad I wasn't clear enough with this issue.
It's not related to the update of the profile on our end, so your changes need to be removed.

We use an external service to show link user profiles and their Twitter. (tzProfiles)
This external service can take a while to propagate, so that's why we would like to display a message down there:

image

"tzProfiles can take a while to synchronize, sometimes up to 1 day. If you've updated your tzProfile recently but the changes are not reflected here, please wait a little bit."

You can use the `text.info` class for the styling of the text

PR #102, opened 2022-04-29
refactor(collections): change algolia search by server search

### LINKED ISSUE #88: Integrate the generic cards explorer header to /marketplace/collections
The "search header" of the page [/marketplace/collection](https://www.fxhash.xyz/marketplace/collections) needs to be updated with the new generic component.

It should be the exact same implementation as on the /explore page. They both request the same data and for now we will use the same filter / sort options.

### LINKED ISSUE #89: Remove algolia from the app
We are no longer using algolia search on the front as it's not being done on the back.

Algolia-related files should be removed to clear the codebase.

* require #88

GemN:
fixes #88 
fixes #89

Few notes:
- After #98 is merged, need to update the button filters for collections
- After #99 is merged, need to update the search minimize for collections

I would like to highlight the files:
- src/utils/filters: get all possible filters value/label in the same place. 
Goal: label consistency, better visibility, dry, in parent component with a search and filters
- src/utils/sort.ts: get all possible sort value/label in the same place.
Goal: label consistency, better visibility, dry, in parent component with a search and sort.
To create a new sort array, we create an array of reference from the sortOptions object.

If it's validated, I can refacto the other files.

Suggestion:
Since the search begin to be more and more complex, yet using the same properties / functions, I would to factorise the search area (filters, sort, and #90).


ciphrd:
Yes, factorizing the filters / sorts will be a great change for clarity and consistency.
Since they all follow the same pattern, I think we can even abstract it one step further by creating an `ExploreParameters` interface which would be inherited whenever we need to plug filter/sort parameters to a CardExplorer.

Something like that:

```ts
interface IExploreParameters {
  sortOptions: IOptions[]
  filterTags: (filters) => ExploreTagDef[]
  // we can probably factorize more into this interface, it's just to have an idea
}
```

I think this should be the scope of a different issue, I'll create it.
I'll keep this issue opened until https://github.com/fxhash/fxhash-website/pull/99 is merged and this PR introduces the appropriate changes to support https://github.com/fxhash/fxhash-website/pull/99

ciphrd (approved):
LGTM

PR #103, opened 2022-05-02
keep iframe history hidden on gentk variation

### LINKED ISSUE #31: Changing <iframe> url pushes a new URL to the browser stack
When on a token page, if we hit the `variations` button a few times, we can only get back to the previous page by hitting the previous button as many times as we clicked on `variations`.

This is actually the expected behavior from browser specs, but is not very good from a UX perspective.

This is classified as a bug, but it truly isn't. We just need to find a workaround.

CiaccoDavide:
fixes #31 

Using `iframe.contentWindow.location.replace(newUrl)` doesn't add the iframe history to the main one!

_Source: https://html.spec.whatwg.org/multipage/iframe-embed-object.html#dom-iframe-contentwindow_

### DIFF
--- a/src/components/Artwork/PreviewIframe.tsx
+++ b/src/components/Artwork/PreviewIframe.tsx
@@ -33,16 +33,17 @@ export const ArtworkIframe = forwardRef<ArtworkIframeRef, Props>(({
   }, [])
 
   const reloadIframe = () => {
-    if (iframeRef.current) {
+    if (url && iframeRef?.current?.contentWindow) {
       setLoading(true)
       setError(false)
-      iframeRef.current.src = iframeRef.current.src
+      iframeRef.current.contentWindow.location.replace(url);
     }
   }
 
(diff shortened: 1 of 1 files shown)


ciphrd (changes requested):
Hey, thanks :) it seems to be a step in the right direction.
It works perfectly fine for the `variations` button, however I noticed the `reload` button now exhibits a strange behavior. The first click doesn't work and shows a blank page, and another clicks show the main page nested. I recorded a video to demonstrate the issue

https://user-images.githubusercontent.com/18054507/166901492-c7fc0b50-4e3b-4aa1-86b4-194031cc4259.mp4

CiaccoDavide:
> however I noticed the `reload` button now exhibits a strange behavior

I don't know how I missed that 😅
I think it's fixed now...I applied the same url-loading method in the reload function

ciphrd (approved):
LGTM, thank you!

PR #104, opened 2022-05-03
Add beta badge for beta tokens

### LINKED ISSUE #64: Add a `BETA` badge on the beta Generative Tokens (field version = 0)

maerzhase:
This PR implements a Badge for Beta Tokens and Objkts. 

- fix #64

# How it looks

### Objkts

Screenshot 2022-05-03 at 18 47 39
Screenshot 2022-05-03 at 18 47 42

### tokens 

Screenshot 2022-05-03 at 18 47 47


src/queries/user.ts
```diff
@@ -62,6 +62,9 @@ export const Qu_userGenTokens = gql`
         thumbnailUri
         labels
         lockEnd
+	objkts {
+	  version
+	}
```
maerzhase, line 67: @ciphrd is there a smarter way to have the version available for the tokens?

src/components/Card/GenerativeTokenCard.tsx
```diff
@@ -30,7 +31,7 @@ export function GenerativeTokenCard({
   lockedUntil,
 }: Props) {
   const url = getGenerativeTokenUrl(token)
-
+  const isBetaToken = token.objkts?.[0]?.version == 0
```
maerzhase, line 34: this is a workaround for the currently query. probably there is a smarter way to add the version to the tokens.

src/queries/user.ts
ciphrd: Damn, I forgot to add a `version` field at the indexer level. We would definitely need to so that this query is optimized before pushing it to prod. It's a small change but requires a re-indexing, so in the meantime I'll keep this PR opened until the changes are made at the indexer's level.

ciphrd (changes requested):
Looking good - We just need a `version` field to be exposed at the indexer's level.
I'll check with the team to see what they think about the design.

maerzhase:
Perfect let me know once we have the version on tokens. i can do the changes quickly then

liamegan:
@ciphrd do we still need this? As we've released and moved into production, the distinction between prod and beta tokens seems to have become much less important. Is there a reason to resurrect this now?

maerzhase:
not having a badge was a product decision.

PR #106, opened 2022-05-05
fix(collab-badge): randomize client-side only to avoid rehydration issues

### LINKED ISSUE #105: BUG: Collab avatars sometimes don't match the artist
Reproduce:
* open project page authored by multiple artists
* refresh
Sometimes the avatars don't match the artist

image
image

We sometimes randomize the order of the artists in a collaboration, this is most likely the root cause of this issue.

GemN:
fix #105 

I encountered this kind of bug in the past, server-side shuffle and client-side shuffle have a different result which mix the results when rehydrating. 
A simple fix is to run the randomisation only on the client-side (using useEffect)

### DIFF
--- a/src/components/User/CollabBadge.module.scss
+++ b/src/components/User/CollabBadge.module.scss
@@ -31,11 +31,11 @@
         font-size: var(--font-size-small);
       }
     }
-  
+
     .avatar {
       border: 3px solid var(--color-white);
     }
-  
+
     .link {
       background-color: var(--color-white);
       border-radius: 100px;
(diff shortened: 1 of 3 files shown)


ciphrd (approved):
LGTM
I still don't quite understand what caused the issue initially though (I understand the disparity between the hydrated state being different from the local one, but I would expect the local one to prevail over the hydrated one).

GemN:
The server-side trigger a random, and get an order in the array, the dom map the array with property key.
When rehydrating, the client-side trigger a random - the order is different, but keys are the same, so it mixes up everything.

PR #116, opened 2022-05-05
feat(activity): change loader component by skeletons blocks

### LINKED ISSUE #110: Activity infinite loader looks bad and should be replaced by placeholder components
Currently there's this ugly loader overlapping with the Actions when Activity feed is loading:

image

We would like to replace this with an optimistic UI (sorry for the Paint sketch):

image

You can reuse the `.placeholder` class in `Effects.module.scss` to achieve the gradient loader effect.

GemN:
fixes #110

### DIFF
--- a/src/components/Activity/Activity.tsx
+++ b/src/components/Activity/Activity.tsx
@@ -3,6 +3,7 @@ import cs from "classnames"
 import { Action as ActionType, TokenActionType } from "../../types/entities/Action"
 import { Action } from "./Action"
 import { useMemo } from "react"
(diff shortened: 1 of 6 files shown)


ciphrd:
We needed the placeholder skeletons to also appear wherever the `<Activity>` component actions are is lazily loaded on scroll, so I made a small refacto to move those skeletons into the `<Activity>` component itself, they only display if a `loading` props is set to true.

We can also leverage the skeleton being placed into the Activity component to have their size match the parent's width.

PR #117, opened 2022-05-06
add 'all time' option on marketplace volume ranking

rsyjswy23:
please let me know if I made it correctly. Thanks a lot.


ciphrd:
Hey !
Thanks for contributing, although you closed your issue I'll give some feedback if you're interested.

Your implementation is correct.

The only issue with this PR is that it introduces a new unnecessary dependency (you may want to update your node version)

PR #119, opened 2022-05-06
Feature: hideable top banner

maerzhase:
This PR implements the closeable top banner with message persistence in the local storage. 
I decided to implement a own context that can be used to stores the messages in the local storage.

Kapture 2022-05-06 at 18 10 51

see comments below


src/context/TopBanner.tsx
```diff
@@ -0,0 +1,73 @@
+import React, {PropsWithChildren, PropsWithoutRef} from 'react';
+import { useClientEffect } from "../utils/hookts"
+
+interface ITopBannerMessage {
+  text: string
+  from: number
+}
```
maerzhase, line 7: Beside the textContent of the message we also store the time when the message was closed. this could be useful for later evaluation of messages, e.g. showing a warning every so and so weeks, etc.

src/context/TopBanner.tsx
```diff
@@ -0,0 +1,73 @@
...
+     ref.current = ctx;
+   };
+
+  const update = (key: keyof ITopBannerContext, value: any) => {
+     const newContext = {
+       ...ref.current,
+       [key]: value,
+     };
+     updateContext(newContext);
+     localStorage.setItem('top-banner-history', JSON.stringify(newContext));
```
maerzhase, line 47: any strong opinions on naming the local storage entity?

src/components/Layout/Warning.tsx
```diff
@@ -1,18 +1,51 @@
...
+      addMessage({
+	text: message,
+	from: Date.now(), 
+      }) 
+    }
+  }
+
+  const messageRef = useCallback(node => {
+    if (node) setMessage(node.textContent)
+  }, [children])
```
maerzhase, line 28: although currently we set the message via `.env` we might want to design more complex banners. therefore the `textContent` is extracted and this is what is stored in the localStorage and checked for.

src/context/TopBanner.tsx
GemN: You can use the src/context/Theme.tsx (which is the global settings context) and add two variables
- hashMessage - save the content of the message as an hash (or not, it's just for easy comparaison later)
- hideTopBanner - save the status of the top banner

whenever the hashMessage is different from the currentMessageHash loaded from the .env variable, you reset hideTopBanner, and update the hashMessage variable

maerzhase: Thanks for the input—Sure lets do this! What do you think about the value extraction of the message value? Currently we are providing span with `dangerouslySetInnerHTML` as the children to the `Warning`. Thats why I am extracing the `textContent` of the node via ref callback. But this now would make it inconsistent to compare the hashed `textContent` with the hashed .env variable. 

I think we should set the type of `Warning.children` to `string` and provide the env variable as string to the `Warning`.  Then we can make a simple comparison against the .env variables. Please let me know what you think

maerzhase:
I changed the implementation:

I removed the additional context and store the message hash in the `SettingsContext`. I introduces the `TopBanner` component that receives a message string. This is also the string that is hashed and compared against. 

let me know hat you think

src/components/TopBanner.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <Warning
+      closeButton={
+	<button
+	  className={cs(style.btn_close)}
+	  onClick={handleClose}
+	>
+	  <i aria-hidden className="far fa-times"/>
+	</button>
+      }
+      className={cs({[style.hidden]: !isOpen})}
```
ciphrd, line 32: The default state should be that the banner is hidden. Because there's a small delay to fetch the local storage, the initial state may not be the one finally computed by the layout once the settings are loaded.
We'd prefer to see a message pop at the top rather to see a weird layout shift to the top when the banner message disappears.
With the current implementation, if we reload the page we may see it shift top and we don't understand why. If we invert this behavior, it makes more sense to see the page shifting down if a message appears at the top.

src/context/Theme.tsx
```diff
@@ -136,9 +140,14 @@ export function SettingsProvider({ children }: PropsWithChildren<{}>) {
     // check for the settings in the local storage
     const fromStorage = localStorage.getItem("settings")
     const values = fromStorage ? JSON.parse(fromStorage) : defaultProperties
+    // set the default for topBannerIsHidden based on the stored hash to avoid bad ux
+    // by shortly showing the banner and then hiding after client side effect
+    const currentTopBannerMessageHash = window.btoa(process.env.NEXT_PUBLIC_BANNER_MESSAGE)
```
ciphrd, line 145: I don't think there's a point in encoding the string with `btoa`, moreover it's not a 100% safe and can throw an error on some utf characters.
I think you can just store the string as it is.

src/context/Theme.tsx
```diff
@@ -149,14 +158,24 @@ export function SettingsProvider({ children }: PropsWithChildren<{}>) {
     const root = document.documentElement
     root.style.setProperty("--cards-border-width", `${context.borderWidthCards}px`)
     root.style.setProperty("--cards-shadow", `${context.shadowCards}px`)
-    root.style.setProperty("--cards-gap", `${context.spaceBetweenCards}px`)
+    root.style.setProperty("--cards-gap", `${context.spaceBetweenCards}px`) 
   }, [context])
 
+  useEffect(() => {
```
ciphrd, line 164: I'm not sure I understand the full-extent of this logic, but if I do I think it is not necessary and can be moved to the initial `useClientEffect`. Maybe i'm missing something though

ciphrd (changes requested):

src/components/Layout/Warning.tsx
```diff
@@ -1,17 +1,28 @@
...
-        <span className={cs(style.message)}>
-          <i aria-hidden className="fa-solid fa-party-horn"/>
-          {children}
-        </span>
-      </a>
-    </Link>
+    <div
+      className={cs(className, style.container)}
+    >
+      {closeButton}
```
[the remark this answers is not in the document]
maerzhase: good point!

src/components/TopBanner.module.scss
```diff
@@ -0,0 +1,31 @@
+@import "../styles/Variables.scss";
+
+.hidden {
+  transition: transform 0.3s ease-in-out, height 0.3s ease-in-out;
```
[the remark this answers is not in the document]
maerzhase: sure np

src/components/TopBanner.tsx
maerzhase: The default value in the store was "hidden". Thats how I wanted to avoid this glitch. anyway's i've updated the implementation: I removed the hidden state from the context because we can store only the message and compute the hidden state from that message within the `TopBanner` itself.

src/context/Theme.tsx
maerzhase: I removed this entirly. We only compute the "hidden" state from the message stored in the store. The state is false by default. I think its cleaner this way

PR #131, opened 2022-05-12
feat(generative-token): add a filter 'listings > all | for sale'

### LINKED ISSUE #128: Add a filter to only see iterations for sale in the project view
In a Generative Token iterations filter, add a filter to filter only the pieces available for sale.
It should use the same component as those available in the explore filters:

image

And propose 2 filters:

- All (default, doesn't apply any filter)
- For sale (apply the filter to only get items for sale)

You can check the marketplace page of a project to see the required graphql filter

GemN:
fixes #128


ciphrd (approved):
LGTM

PR #133, opened 2022-05-13
feat(user): replace tab activity by dashboard and add sales component

### LINKED ISSUE #122: Replace user activity by user dashboard
We want to replace the `activity` tab in the user profile by a more generic-purpose dashboard in which we'll centralize a few features to make it easier for users to inspect their activity as well as the activity of other users.

Here's a design for the implementation of the first iteration of the dashboard feature:

Dashboard (1)

The first iteration of the dashboard will only have a `sales` table in which we will put the user sales on the secondary market, either as the direct seller or as the author of a piece. The sales endpoint will cover all these cases.

- [ ] replace `activity` tab by `dashboard` tab (and update the URL too)
- [ ] create a 301 redirection from `activity` to `dashboard` url in the `next.config.js` file
- [ ] the current activity infinite scroll should be moved at the bottom of the page
- [ ] call the `user { sales }` endpoint to query the sales and display the sales in the table
- [ ] handle responsiveness of the table

Here's some implementation details layered on top of the design:

Dashboard (2)

For the implementation of this feature, the API will need the `sales` field in the `user` to be implemented:

* https://github.com/fxhash/fxhash-api/issues/18

GemN:
fixes #122 

Missing from API (for user { sales }):
- Sort by createdAt-desc
- Filter by type in [LISTING_V1_ACCEPTED, LISTING_V2_ACCEPTED]
- Pagination stuff for infinite scroll

Need to review:
- Responsive, what to display to mobile user ? (currently, only name and price that I think are the most importants cols)
Do we want to display different information at a glance?
Do we want to display more information on column tap ? (or 3 vertical dots at the end of the row)

Capture d’écran 2022-05-13 à 14 32 35

New implementations: 
- It's the first time next/image is used in the project, I'm not sure if it's deliberately avoided - This next component serve the right size of image (for static and DYNAMIC urls) and allow some additional features (avoid layout shift, image fit, srcset, blur placeholder etc), so I had to whitelist  `domains: ['gateway.fxhash.xyz', 'gateway.fxhash2.xyz']` in next.config.js 
- There are two components in TableUserSales, one for desktop, one for mobile. It is to avoid the DOM rendering / calculating unnecessary columns with display: none. (shout me a DM if you want to know more)

Suggestions for the future:
Sales list can be long and a waste of performance for the DOM to display everything even tho we don't see it.
In a second version, it would be a good idea to implement table virtualization.


ciphrd (changes requested):
I updated the API endpoint:
- now sorts by date
- supports pagination

I made a few changes to the styling. Regarding the mobile version, I think we can use the same table but could make it scrollable on the x axis. This way no data is lost. For this feature, I think you can implement a min-width on the table so that it's responsive over such a width, but scrollable under it.

image
_x-scrollable table in the doc, mobile (https://www.fxhash.xyz/doc/fxhash/integration-guide)_

I moved the price column as the 2nd column, this way the css mobile might be configured so that the token name & price are displayed when landing on the page.

Also a few remarks regarding the table layout:

image

image

The pagination can be triggered when we get near the end of the table. I've implemented pagination into the graph QL query and in the apollo cache, you'll just have to implement the trigger of the query.

GemN:
Ready to be review

ciphrd:
LGTM

PR #138, opened 2022-05-19
Add scoped cardsize select to more views and load HQ preview when on large cardsize

### LINKED ISSUE #130: Add the size select to the `/explore`, `/incoming`, `/marketplace`, `/marketplace/generative/{id}` pages
Add the size select component to the given pages

image

Because we have 2 resolutions available, we need to make sure that if the size is set to big on these pages, the HQ image is used instead of the LQ one.

- [x] `/explore`
- [x] `/incoming`
- [x] `/marketplace`
- [x] `/marketplace/generative/{id}`
- [x] use the HQ preview if the size is set to big

`/explore` and `/incoming` should have the same scope-identifier.

@maerzhase maybe you'll want to take this one since you've been working on this feature.

maerzhase:
- fix #130 

This PR implements caching preview urls by key in the `Card` component. This allows us to switch the previewUrl to HQ url when large card size is selected. 

@ciphrd As I already mentioned `CardExplorer` on retained routes, caused a race condition when updating the `--card-size` css variable. All retained `CardExplorer` views would receive the pathname change which would cause in inconsistent card sizes. I could fix that by sorting the retained routes by visiblity, so the visible component is the last that triggers it effects.

Shall we had the card size selection also to the `u/<name>/creations` and `u/<name>/collection` pages?


maerzhase:
merge conflicts solved. ready for review

ciphrd (changes requested):
I noticed that the card size feature is not working on a project iterations:

image

maerzhase:
good catch gonna have a look

ciphrd (changes requested):
Looking good!

Only detail: the setting is missing from the user collection, I think it'd be great in there

GemN (approved):
LGTM! Good job :)

GemN:
Actually I found a bug (or undocumented feature?) 
Open filters in iterations with "big card size" 
Capture d’écran 2022-11-15 à 18 52 20

maerzhase:
thanks @GemN !! good issue you found and i fixed another leek we discussed!!

GemN (approved):

src/components/Exploration/CardsExplorer.tsx
```diff
@@ -52,20 +55,33 @@ export function CardsExplorer({
...
+
   // get cardSize from scope or use default
   const cardSize = useMemo<number>(
-    () => (!cardSizeScope ? DEFAULT_SIZE : settings.cardSize),
-    [settings.cardSize, cardSizeScope]
+    () =>
+      !cardSizeScope || !settings.cardSize[cardSizeScope]
+        ? DEFAULT_SIZE
+        : settings.cardSize[cardSizeScope],
+    [cardSizeScope, settings.cardSize]
```
maerzhase, line 65: Since card size is now provided by scope we check if we have a card size saved or use default size

src/containers/User/UserGenerativeTokens.tsx
```diff
@@ -60,7 +60,7 @@ export function UserGenerativeTokens({
         onTrigger={handleFetchMore}
         canTrigger={!loading && !hasNothingToFetch}
       >
-        <CardsContainer>
+        <CardsContainer cardSize={270}>
```
maerzhase, line 63: Since user creations don't have the cards explroerer we don't show card size select and provide the default card-size to the container

Issue #70, opened 2022-04-23
Masonry layout in the iterations view


ciphrd:
On the Generative Token page > iterations, the grid looks weird when the images don't have the same dimensions (lots of white space).
A masonry layout would be more appropriate in such instances.

This project can be used as an example:
https://www.fxhash.xyz/generative/slug/(kinder)garden-monuments

There are some aspects to take into account to implement this feature:
* a new `CardsMasonryContainer` component should be created as we cannot use the current one
* no javascript to place elements with absolute values
* the order of the cards must be preserved
* a refacto of the infinite scroll trigger should be implemented so that we refetch whenever we meet empty space at the bottom of a column
* the number of columns should match the current grid column size css implementation (window resize listener too)

Cerfio:
Good luck to me, I'm on it

maerzhase:
i opened a PR based on the existing masonry PR .

liamegan:
@ciphrd there's a stale PR for this, see https://github.com/fxhash/fxhash-website/pull/153 we should probably update this and get it merged if we can.

PR #153, opened 2022-05-21
Add Masonry Component with dynamic cardSize support

maerzhase:
This PR is based on:
- #59 thanks @[jbmoutout](https://github.com/jbmoutout)[jbmoutout](https://github.com/jbmoutout)

and fixes:
- #70

______________________________

Regarding this: 
> ⚠️ issue between the CardLoading logic and masonry when replacing placeholder height (see loom) that would require more work. https://www.loom.com/share/19388a89e654497187473553e04a822b

I cannot reproduce it or understand what is meant here. 

@ciphrd i merged dev and also included the support for dynamic `cardSize`. 
Kapture 2022-05-21 at 19 35 14


ciphrd:
I will ask a review from the team on this, after taking another look the layout looks great.

On a side note, I was reviewing #138 just before and I noticed an issue with the cardSize being an object.

image
image

Probably related to the issue reported in #138, and something to keep an eye on if we were to merge the branches.

ciphrd (changes requested):
The implementation is great and the feature works well.

As pointed our by the team, it would be great to have an option to toggle the masonry layout.

I'm thinking about creating a new category "Layout" in the settings, in which we would put this setting.

image

I think it makes sense to have another category because it's more of a container rule than a card design rule. As such, the `Gap between cards` setting could be moved to this new category.

ciphrd (approved):
LGTM

Made a few changes due to a bug where the cards loaded were moved our of the Cards Container, breaking the placeholders effect.

PR #154, opened 2022-05-24
fixed reserveProgress calculation

CiaccoDavide:
20220524_080956_firefox_uHgw5JhO9H

20220524_080621_firefox_GZ54qUZJUs

reserve progress width was wrong when the burnt edition weren't shown

solves #155


CiaccoDavide:
I also added `Frag_GenReserves` to `Qu_genTokenMarketplace`, so that now the reserves bar is shown also in the `marketplace/generative/[id]` page

ciphrd (approved):
LGTM

PR #158, opened 2022-05-26
style(project): ignore IDEA files from intellij idea

GemN:
A small change to .gitignore to avoid pushing IDE-related files from Intellij Idea to the repo (IDE that I use)


liamegan:
@ciphrd please review

PR #166, opened 2022-05-27
feat(user): add sub tabs to dashboard (sales, offers received, offers sent, activity)

### LINKED ISSUE #149: Create sub-tabs in the user dashboard
The PR should be done against the `feature/offers` branch.

- [x] Create a new tabulation component from the given design
- [x] Move the sales table user under sales tab and adjust its design to match with the new one (basically change the header color to black)
- [x] create 2 tabs, one of for `offers (sent)` and one for `offers (received)`, leave those tabs empty

Image

GemN:
fixes #149

- deeplinked tabs
- improved tabs component to accept key as active index
- add a new style to tabs component ("subtabs" feel free to rename)
- add scroll: false behavior for smooth navigation


ciphrd:
LGTM
I moved the dashboard as the last tab element as it makes more sense to showcase artist pieces through their profile URL

Issue #32, opened 2022-01-07
Fix sandbox service worker not working on firefox


ciphrd:
The sandbox uses a service worker to fake the response from browser network calls made within the sandbox iframe.
It works fine in chrome, but doesn't in firefox.

PR #173, opened 2022-05-30
Sandbox: Fix file fetch() security issue in serviceworker (#32)

L-A:
I've decided to dig into the reason why Firefox fails to retrieve URLs in the sandbox's iframe context (#32). This is the PR that fixes it 🎉 

Explanation:

The only immediate symptom is network requests failing with `NS_ERROR_MALFORMED_URI`, which remains pretty opaque when it happens for URLs like `/sandbox/somefile.js`. That points to the request or the response content being non-compliant in some way related to networking.

Starting from the source: the logic itself in `worker.js` is sound. In fact, the worker _sends_ a `Response` to the iframe, even in Firefox. What's happening is that this response then fails something like a CSP check, and is discarded.

So why? It turns out that the iframe context is under tighter content security than the service worker's. **It's possible for a request fired from the service worker to return a Response that is too permissive to be forwarded whole to an iframe**. In this situation, it must be rejected and the original request ends up failing.

That's what happens in Firefox. I suspect that, given enough time, other major browsers will adopt the same policy.

Fortunately, the solution is simple*: Instead of forwarding the `Fetch` request initiated from the service worker's context, we build a new `Response`, which will have a permission level that's compliant with the request it's responding to.

[**The Response `type` policy is explained here**](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent/respondWith). 

_* says me, who spent a couple of afternoons instrumenting everything until I figured out the precise cause_


ciphrd (approved):
LGTM !

Thanks **a lot** for investigating into this somehow haunted issue, and for finding the root cause 🎉 .  As you mentioned the Firefox error is not very explicit so thanks for digging.

PR #174, opened 2022-05-30
Fix MessageCenter wouldn't remove messages

### LINKED ISSUE #169: Notifications cannot be cleared by clicking the X
Clicking the `X` on a notification doesn't clear it

Image

### LINKED ISSUE #170: Notification container remains although notifications disappeared
When notifications are hidden, the container remains.

This is most likely related to #169 and it seems like the "hide" event is triggered (hiding the notifications with a slide) but the clear event doesn't work.

Image

maerzhase:
The fix was using the actual memoizedContext that would have the correct removeMessage handler. 

- fix #170
- fix #169


ciphrd (approved):
LGTM !

PR #175, opened 2022-05-30
Use EntityBadge on gentk page

### LINKED ISSUE #171: [mobile] Gentk page is not using the EntityBadge component
The Gentk page doesn't show the authors of a collab on the mobile header

Image

maerzhase:
- fix #171 

@ciphrd I had to add a `centered` prop to the `CollabBadge` which was necessary because the `UserBadges` on the gentk page have the `prependText="created by"` and therefore create a strange layout. Wasn't an issue before because other badges on other pages don't have the `prependText`.


src/components/User/CollabBadge.tsx
```diff
@@ -11,12 +11,14 @@ import { getUserName, isUserVerified } from "../../utils/user"
 
 interface Props extends IEntityBadgeProps {
   user: Collaboration
+  centered: boolean
```
ciphrd, line 14: I believe the centered prop should be optional, otherwise there will be compilation errors

ciphrd (changes requested):

src/components/User/CollabBadge.tsx
maerzhase: dang pushed the fix

ciphrd (approved):
LGTM!

PR #181, opened 2022-06-02
fix(user): moderation tooltip display correctly and doesnt destroy layout

### LINKED ISSUE #179: Tooltip in UserHeader clipped
The Tooltip in the UserHeader is currently clipping because of the overflow: hidden 

Screenshot 2022-06-01 at 17 43 17

@GemN this was introduced as a fix for mobile maybe we should only add this style on smaller breakpoints?

GemN:
fixes #179


maerzhase (approved):
finally! 👍 🐰

PR #189, opened 2022-06-09
feat: markdown parser with math, embed media widgets, custom elements

### LINKED ISSUE #182: Extended Github Flavored Markdown viewer
See #180 for general context.

This issue aims at building a renderer for the extended Github Flavored Markdown (eGFM) described in the following article: https://www.notion.so/fxhash/Article-content-spec-5fe1029e69b748b1beac36db3b3f20d7

We need a general GFM parser which turns the markdown into a node tree (we can probably reuse https://github.com/fxhash/fxhash-website/blob/3e61eb987af98fe44356b952d9b23bd063a291d9/src/services/LocalFiles.ts#L153 without the html parser).

We also need a parser which can identify the custom syntax proposed in the article, and turn it into a node.

Finally, a renderer to process the nodes using React components.

- [x] GFM parser
- [x] extended syntax parser
- [x] node tree renderer

The PR should be made against the `feature/markdown-editor-viewer` branch. We can use temp pages & test data to facilitate testing & iterating.

GemN:
fixes #182


ciphrd (approved):
LGTM, merging this to move forward

PR #205, opened 2022-06-16
feat(article): add article page

GemN:
require merge #204 
fixes #192

PR #206, opened 2022-06-16
feat(article): add related article design

GemN:
ongoing #193

require #205 

_(I'll copy what I put on discord to keep it within the scope)_ 

for #193 to query related articles, that would be the best to have 
```graphql
query {
  article(id) {
    ...
    relatedArticles(limit: 4) {
      ...
    }
  }
}
```
And on the back-end to query similar articles by tags:
get all articles order by
- articles that match mosts of the tags (and at least one)
- created at (or closest to current article date if you want to avoid repetition for articles sharing the same tags => most recent will be "advantaged")

At the moment, that require to first fetch the article tags THEN make a second query for the related


src/containers/Article/ArticleInfos.module.scss
```diff
@@ -0,0 +1,73 @@
...
+  margin-top: $spacing;
+  display: grid;
+  grid-template-columns: max-content 1fr;
+  align-items: baseline;
+  gap: 12px 5px;
+  /* prevents children elements from streching */
+  justify-items: start;
+  // & > *:nth-child(2n):not(a) {
+  //   opacity: 0.6;
+  // }
```
bikkimahato, line 63: Commented code should not be pushed to GitHub.

GemN:
fixes #193

PR #238, opened 2022-06-26
implement floating inline menu for text formats

maerzhase:
This PR Implements the `FloatingInlineMenu`. For now it can only set inline styles but we can iterate on it to add other buttons that format blocks etc.


maerzhase:
there is an issue with the block level menu that we still need to resolve before merging

PR #247, opened 2022-07-07
Fix image to figure processing

maerzhase:
require #245 

This PR is based on PR #245 and fixes an issue that @GemN found when images processed from markdown to slate editor. 

The fix happens in two stages:
- Since images are inline in markdown they would end up having a paragraph wrapped around it. `remark-unwrap-images` is used to removes the wrapping paragraph of images
- Then the markdown images are processed into the desired figure structure


src/components/NFTArticle/processor.ts
```diff
@@ -159,12 +161,31 @@ function createMathNode(node: any) {
...
+        text: ""
+      }],
+    }, {
+      type: "figcaption",
+      children: [{
+        text: node.alt
+      }]
+    }]
+  }
+}
```
maerzhase, line 180: here we transform the markdown image into the structure a figure expects.

PR #250, opened 2022-07-07
keep focus on editor when adding blocks

### LINKED ISSUE #230: Article editor, focus new elements when they're created

When a new element is created using the `+` button, the focus is lost. The text-editeable area should be automatically focused when possible (each block may require a particular treatment if needed).

https://user-images.githubusercontent.com/18054507/175791829-ef5df7cd-ded8-42eb-b22a-9bf592dd6cd4.mp4

maerzhase:
~~pointer down events are used instead of click to
prevent the cursor loosing focus in the contenteditable
pointer event is used for compatibility with different
input devices. when using add block the selection
is set to the add block target path.~~

We dont use pointerDown, but found a way to restore the selection after the focus is lost. see comments below

- fix #230

### DIFF
--- a/src/components/NFTArticle/SlateEditor/Elements/RenderElements.tsx
+++ b/src/components/NFTArticle/SlateEditor/Elements/RenderElements.tsx
@@ -48,14 +48,10 @@ function EditableElementWrapper({
     // in order to retrieve the DOMNode and restore
     // the selection correctly, we have to wait
     setTimeout(() => {
-      const domNode = ReactEditor.toDOMNode(editor, element)
-      function setSelectionToNewBlock() {
-	ReactEditor.focus(editor)
-	const path = ReactEditor.findPath(editor, element)
-	const [lastLeaf, lastLeafPath] = Node.last(editor, path);
-	Transforms.select(editor, lastLeafPath)
-      }
-      domNode.tabIndex > -1 ? domNode.focus?.() : setSelectionToNewBlock()
+      ReactEditor.focus(editor)
+      const path = ReactEditor.findPath(editor, element)
+      const [, lastLeafPath] = Node.last(editor, path);
+      Transforms.select(editor, lastLeafPath)
     })
   }
 


src/components/NFTArticle/SlateEditor/Elements/RenderElements.tsx
```diff
@@ -45,6 +45,19 @@ function EditableElementWrapper({
...
+      const node = Node.get(editor, target)
+      const domNode = ReactEditor.toDOMNode(editor, node)
+      function setSelectionToNewBlock() {
+	ReactEditor.focus(editor)
+	const path = ReactEditor.findPath(editor, node)
+	const [lastLeaf, lastLeafPath] = Node.last(editor, path);
+	Transforms.select(editor, lastLeafPath)
+      }
+      domNode.tabIndex > -1 ? domNode.focus?.() : setSelectionToNewBlock()
+    })
```
maerzhase, line 60: After a block is added, we check if the created element is focusable. If `true` we call the elements `focus()` method. If `false` we restore the selection by setting the focus to the editor and finding the deepest leaf path of the created node and set the selection there.

src/components/NFTArticle/elements/Figure.tsx
```diff
@@ -17,12 +17,22 @@ export function FigureElement({
...
   const focused = useFocused()
+  const editor = useSlateStatic();
+
+  const handleFocus = () => {
+    const slateNode = ReactEditor.toSlateNode(editor, attributes.ref.current);
+    const path = ReactEditor.findPath(editor, slateNode)
+    const captionText = Node.string(slateNode)
+    const point = {path: [...path, 1], offset: captionText.length }
+    Transforms.select(editor, point)
+  }
```
maerzhase, line 28: When a figure is focused. We want to restore selection to the caption inside the figure.

ciphrd (changes requested):
It seems to be working well for paragraphs.
However, when clicking an image there is a weird glitch and the focus on the caption doesn't seem to be working anymore (see video attached).

https://user-images.githubusercontent.com/18054507/180054788-27622a25-c92a-4fff-892c-71116bbfcc6c.mp4

Please also make sure it's working with all the blocks, I didn't check.

maerzhase:
@ciphrd turns out that we must ensure to render the `children` that are provided by slate, also if the component is e.g. avoid component. This allows us to create a proper slate selection and the we can actually just use `useSelected` hook from `slate-react` to determine if a block is selected and trigger and effect to e.g. focus the textarea. I've implemented it for the KatexBlock. 

please note: the media embed is not yet implemented on this branch.

src/components/NFTArticle/SlateEditor/Elements/Blocks.tsx
```diff
@@ -316,10 +316,9 @@ export const BlockDefinitions: Record<EArticleBlocks, IArticleBlockDefinition> =
     icon: <i className="fa-solid fa-function" aria-hidden/>,
     buttonInstantiable: true,
     render: ({ attributes, element, children }) => (
-      <div className={style.article_wrapper_container}>
-        <BlockKatexEditor slateAttributes={attributes} slateElement={element}>
-          {children}
-        </BlockKatexEditor>
+      <div className={style.article_wrapper_container} {...attributes}>
+	{children}
```
maerzhase, line 320: we must ensure that we render the `children` otherwise will throw an error when the selection is pointing to this block

ciphrd (approved):
LGTM

PR #253, opened 2022-07-08
feat(article): add math editor component with katex

### LINKED ISSUE #236: Article editor, add support for the math block

The math block is composed of katex render and a text input at the bottom of the math render. When the block isn't focused, only the katex render is displayed. When the block is focused, the text input is displayed and can be edited.

If the text input is empty, the katex render is hidden and the text input is displayed in all circumstances. The text input has a placeholder "Enter a Latex math formula".

GemN:
require PR #245
fixes #236


maerzhase (approved):
This works really great! good job!

ciphrd (changes requested):
Good job, only one remark: when the block is unfocused, the input text should be hidden. When clicking on the math formula, we should show the input again.
The idea is to display the article as it's going to look in the end.

ciphrd (approved):
LGTM

PR #256, opened 2022-07-18
feature/articles Allow inline styles and links in image caption

### LINKED ISSUE #231: Article editor, formatted text causes caption fragmentation after markdown export when used in figcaption

# Steps to reproduce

* drop an image in the editor
* click the image, enter a caption at the bottom
* select a section of the caption
* press `CTRL` + `i`
* click on `preview & mint` at the bottom, which logs the markdown to the console

# Demo

Image

```markdown
![Some text, ](blob:http://localhost:3000/aea137a2-6cfc-4c21-8bdb-e09990cf05f2)
```

The italic node cuts the text in the caption

# Probable location

https://github.com/fxhash/fxhash-website/blob/f2b4294f7ea85d9a6f74ae8fd6f7651e5d132bc3/src/components/NFTArticle/processor.ts#L219

# Solution

- [x] remove text styling on the figcaption (no menu, no shortcut)
- [x] enforce figcaptions to only be a simple text node through normalization (https://docs.slatejs.org/concepts/11-normalizing)

maerzhase:
This PR implements inline styles and links for image figcaption. I used the `title` attribute (instead of alt) of the markdown image for serialising the inline styles into a md string. The `alt` attribute should be a plain string because the alt attribute should also be used for a11y, it shouldn't contain inline styles or links. 

- fix #231

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "is-hotkey": "^0.2.0",
     "is-mobile": "^3.0.0",
     "lodash.throttle": "^4.1.1",
+    "mdast-util-from-markdown": "^1.2.0",
     "nanoid": "^4.0.0",
     "next": "12.0.7",
     "nextjs-progressbar": "^0.0.13",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "is-mobile": "^3.0.0",
     "lodash.throttle": "^4.1.1",
     "mdast-util-from-markdown": "^1.2.0",
+    "mdast-util-to-markdown": "^1.3.0",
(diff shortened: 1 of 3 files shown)


ciphrd (changes requested):
I'm not sure how you implemented this, but given your description it seems my issue was probably incorrectly formulated.
For fixing the bug, I would suggest simply removing the text formatting for the caption since the image alt doesn't support it.

Moreover, some normalization on the alt property should be added to enforce only a simple text node there: https://docs.slatejs.org/concepts/11-normalizing

PR #257, opened 2022-07-22
feature(article): add table

### LINKED ISSUE #234: Article editor, add support for tables

Tables will always have a head by default.
Add/remove rows & cols using buttons on the sides/at the bottom of the table, a little bit like this:

Image

The buttons should only be visible if the table is focused.

Add table support for slate to markdown.

GemN:
fix #234


GemN:
@ciphrd it's ready to be review - a note, before the merge we would need to change the icon for deletion of row / column

PR #260, opened 2022-07-25
Fix/#231 normalize figure alt text

### LINKED ISSUE #231: Article editor, formatted text causes caption fragmentation after markdown export when used in figcaption

maerzhase:
__This PR requires #250__

@ciphrd I introduced a new boolean `hideFloatingInlineMenu` to the `BlockDefinitions`. This way we can control which nodes will actually show the inline style menu. Further more i've also added the normalisation of the figcaption into the `ImagePlugin`. 

- fix #231


ciphrd (changes requested):
Great job!

One remark: it's still possible to paste an image inside the figcaption, the normalization should also prevent that (and not only that but also any non-text leafs)

image

src/components/NFTArticle/SlateEditor/Plugins/SlateImagePlugin.ts
```diff
@@ -88,7 +103,13 @@ export const withImages = (
...
             })
           }
-        }
+	} else {
+	  // any child that is not the first image in the figure moved after the figure
+	  if (!(child.type === 'image' && childPath[childPath.length - 1] === 0)) {
+	    Transforms.liftNodes(editor, {at: childPath})
+	    return;
+	  }
+	}
```
maerzhase, line 112: @ciphrd I added this line which moves any child beside the first image after the figure. We could also change the `liftNodes` to `removeNodes` but imho the UX is better when e.g. the images are actually pasted into the editor. What do you say?

ciphrd: Yes I believe this is a great UX, perfect :)

ciphrd (approved):
LGTM!

Merging

PR #263, opened 2022-07-26
Fix/#232 add shortcut for creating links and allow configuration of inline menu via blockDefinitions

### LINKED ISSUE #232: Article editor, add shortcut to insert a link
* requires #217 

Once the popin to manipulate some text will be done, there should be a way to turn some text into a link. The same feature should be triggered by the `CTRL` + `K` keyboard shortcut.

maerzhase:
As mentioned I've removed the hotkey plugin in order to add hotkey functionality to the inline toolbar itself. It also allows to activate formats and keep typing in the activated format without selecting a range of text.

- fix #232


maerzhase:
@ciphrd this one is ready I it helps us to define the behaviour of the inline toolbar within the current selected block and the approach of the link can be used to create some form for the inline math

ciphrd (approved):
LGTM

PR #287, opened 2022-07-28
feat(articles): normalisation of tables (handle bad formatted table and deletion of cells)

GemN:


ciphrd (changes requested):
Looking at the GFM spec, I think we should only allow for leaf type nodes inside the cells:
https://github.github.com/gfm/#tables-extension-

> A [table](https://github.github.com/gfm/#table) is an arrangement of data with rows and columns, consisting of a single header row, a [delimiter row](https://github.github.com/gfm/#delimiter-row) separating the header from the data, and zero or more data rows.

> Each row consists of cells containing arbitrary text, in which [inlines](https://github.github.com/gfm/#inline) are parsed, separated by pipes (|). A leading and trailing pipe is also recommended for clarity of reading, and if there’s otherwise parsing ambiguity. Spaces between pipes and cell content are trimmed. Block-level elements cannot be inserted in a table.

Right now it's possible to paste container nodes in the cells for instance: 
image

I'll leave this note here as I believe it makes sense to process it here, but eventually we can make into another issue

GemN:
Yes, I agree we talked about that with @maerzhase we believe the best solution would be to have a/few parameter/s to blockElements that indicate if inside a X element we can:
- use inline menu
- what can we use inline
- use the block menu
- what can we use from the block menu

GemN:
Temporary fix that allow only text in tables when there is multiple fragments.
 I limited the inline menu to have only bold/italic in table cells

ciphrd (approved):
LGTM

PR #289, opened 2022-07-28
feat(article): add tags to article; fix(article): fix display image in preview display correctly

### LINKED ISSUE #261: Article editor, add tags input
* target PR branch: `feature/articles`

Add a tags full width text input with info text "A list of comma-separated values" before the number of editions.

image

The tags should be auto saved like the other fields, and eventually sent to the backend at the time of the upload.

GemN:
fix #261


ciphrd (approved):
LGTM

PR #293, opened 2022-07-29
feat(article): unwrap listItem for slate

### LINKED ISSUE #281: [Article editor] Fix lists conversion to markdown (creates multiline)
* PR target: `feature/articles`

Steps to reproduce:
* create article
* add a list
* add a few points
* wait for save
* reload:

image

It appears that the bug occurs at the slate to markdown conversion, this is an output of the markdown locally saved:
image

However, it could also be an issue at the edition stage (pressing enter could ne the cuase, needs investigation)

As you can see, it adds an empty line between each bullet, which tricks slate into create multiple lists inside paragraphs.

- [ ] fix the conversion issue (no empty line between list points)
- [ ] add normalization rule so that <ul> elements must be at the root

GemN:
fixes #281


ciphrd (approved):
LGTM

I also added the `spread: false` option on instanciation to prevent new lines between every item, which was breaking parsing

Issue #284, opened 2022-07-28
[Articles] Add search & sort


ciphrd:
* base branch `feature/articles`

It's currently not possible to search for articles.
The same filter component can be reused: 

![image](https://user-images.githubusercontent.com/18054507/181525891-65c4a1e8-96bc-44fe-8a45-9e2b0bcea886.png)

Filters:
- [x] search string (through the search bar)

Sort:
- [x] created at, ASC DESC
- [x] relevance (when search active, ASC DESC)

It's already implemented on the API, same interface than Generative Tokens

PR #295, opened 2022-07-29
feat(articles): add sort and search to article

GemN:
fixes fxhash/fxhash-website#284

also add refacto to sort and filters


ciphrd (approved):
LGTM

Nice refacto

PR #306, opened 2022-08-01
Feature/artciles: add softbreak plugin

### LINKED ISSUE #304: [Article editor] `SHIFT` + `ENTER` should insert a new line, and not insert a new block
* base branch: `feature/articles`

When the cursor is in the following blocks:

- p
- ul/ol
- blockquote

Pressing `SHIFT` + `ENTER` should insert a new line within the block

maerzhase:
-  fix #304


GemN:
I think it's too permissive at the moment, you're not checking the node type, it could add undesired side effects 
```js
  editor.insertSoftBreak = () => {
    // check the node type is p / ul / ol etc
    // insert \n
    insertSoftBreak() // to allow other plugins to work if they manipulate soft break
  }
```

maerzhase:
sure we can add that, i am not sure if its necessary though. softlinebreak is only triggered on text nodes. that should already exclude all custom blocks. 

but we can also retrieve the node by cursor position and do a check for the type.

maerzhase:
I included the changes proposed by @GemN 👍

ciphrd (changes requested):
I believe it's good to check for the block type, as we want to avoid any weird side effect.

There is an issue in the markdown transformation, this is the string which gets computed:

```md
This is a test&#x20;
We can have multiline now, great

Another test
```

(see here https://gateway.fxhash-dev2.xyz/ipfs/Qmcfk4M1jwkFiN3FhEw1m1PoZX5nqjSPKqGHxuFKAM7qgV)

Resulting in the viewer not displaying the multiline properly:

image

I believe the resulting markdown should be:

```md
This is a test
We can have multiline now, great

Another test
```

This is by default rendered like that by markdown viewers:
> This is a test We can have multiline now, great

As there is no support for backline in paragraphs. It needs to be added.

maerzhase:
the encoding comes from the mdast-util-to-markdown: https://github.com/syntax-tree/mdast-util-to-markdown/commit/108450c5e3dc013f8e9f7119b63ed8feddff7a7b

when i started to replace the encoded whitespaces it also breaks the newlines. @ciphrd i am not sure how to resolve this issue

PR #312, opened 2022-08-03
use bulletOther option to avoid comments between lists

### LINKED ISSUE #296: [Article editor] Slate to markdown introduces a weird html comment between 2 list blocks
* base branch: `feature/articles`

It's a very odd behavior.
When adding 2 lists one after the other, the parser adds this weird comment between the 2 list blocks ?!:

image

It only happens when we add 2 lists next to each other, weird af.

maerzhase:
- fix #296

markdown string looks like this when using bulletOther option:
Screenshot 2022-08-03 at 20 40 22


ciphrd (approved):
LGTM

PR #313, opened 2022-08-04
feat(article): insert paragraph after break for heading

### LINKED ISSUE #279: [Article editor] Pressing `ENTER` at the end of a heading should create a paragraph
* PR target: `feature/articles`

Right now, when we press `ENTER` at the end of a title it create another title block above which rom a UX standpoint, is not good.

GemN:
fix #279

I made sure we could define the break behavior by block definition, might be borderline YAGNI but I'd like the end goal to have only one main file to modify when we add a new block.

This PR also modify the behavior when editing a list, if the last list item is empty, it's transformed as a new paragraph. (behavior similar to notion)


ciphrd (approved):
LGTM, great implementation

I believe it's great to ship s little more code if it brings a modular and upgradeable implementation.

Issue #283, opened 2022-07-27
[Articles] Bundle optimization


ciphrd:
* base branch: `feature/articles`

Bundling with the articles in their current state yields quite a big bundle, we need to optimize it (+ 1MB, too much)
Maybe there's a way to only use slate on the backend when needed (for instance displaying an article), but it needs further investigation.

Branch `feature/articles` at 564b9f9c63faa3f81f0e121cc3595ac1ae85b43f

```
Page                                         Size     First Load JS
┌ λ /                                        3.43 kB         690 kB
├   └ css/7af24baffdc14fc6.css               1.85 kB
├   /_app                                    0 B             687 kB
├ ○ /404                                     843 B           688 kB
├ ○ /500                                     865 B           688 kB
├ λ /article/[slug]                          373 B          1.05 MB   <--- IMPORTANT
├ ○ /article/editor/local/[id] (435 ms)      855 B          1.11 MB   <---
├ λ /article/editor/local/[id]/preview       5.22 kB        1.05 MB   <---
├   └ css/7d93509a296788f6.css               323 B
├ ○ /article/editor/local/new                869 B          1.11 MB   <---
├ λ /article/id/[id]                         371 B          1.05 MB   <--- IMPORTANT
├ ● /articles/[id] (662 ms)                  1.4 kB          688 kB
├   └ css/7f87720026426dad.css               1.42 kB
├   ├ /articles/about-fxhash
├   ├ /articles/beta
├   ├ /articles/changelog
├   └ [+8 more paths]
├ ○ /collaborations                          3.99 kB         698 kB
├   └ css/fd08c8190b64f54f.css               1.42 kB
├ λ /collaborations/[address]                11.5 kB         740 kB
├   └ css/df9dbb4237ab4764.css               3.7 kB
├ ○ /community/opening-schedule              28.2 kB         719 kB
├   └ css/0ba012f8c9c76efd.css               1.75 kB
├ λ /community/reports                       5.15 kB         692 kB
├   └ css/c6d373d175a7b8cd.css               1.3 kB
├ ● /doc                                     1.61 kB         689 kB
├   └ css/9f4d21fd0bb12f98.css               1.25 kB
├ ● /doc/[category]/[article] (5731 ms)      2.15 kB         693 kB
├   └ css/514ee53ec20dbdcb.css               1.35 kB
├   ├ /doc/artist/reserves (600 ms)
├   ├ /doc/artist/code-of-conduct (592 ms)
[...]

PR #336, opened 2022-08-16
refactor(article): simple folder structure, add guidelines to create plugins and new elements

### LINKED ISSUE #278: Cleanup the architecture & file/folde structure of `NFTArticles`
We lacked some guidelines when iterating on the implementation of the articles, we may need to refacto the architecture a little bit & the file/folder structure as components sharing similar properties are dispatched at different places in the folders.

GemN:
require PR #335 (I started the refacto before, there and there)

fix #278 

I added a README.md that explains the folder structure and how to add features.

Basically, all data related to one element (for ex: Heading) is grouped in one folder.
In that folder you'll find key files:
- ElementDefinition.tsx : mandatory, contains all the data related to the element behavior in the editor
- ElementEditor.tsx: optional - component used in the render prop of the element definition, it's optional because some elements don't necessarily need a component. (ex: paragraph)  
- ElementDisplay: optional - used to display the element outside the editor (sometimes the processor do it for you, that's why it's optional)
- ElementProcessor: optional - used for plugins in processor, when a custom component need transformation in processor, the function is defined here

Best way to review this PR, is to fetch the branch and check the NFTArticle folder (there is a lot of moved files / rename)


GemN:
Now Build correctly without error (for some reasons Prism throw error with SSR)

I also fixed bundle size first load js. Basically dynamic load unified save 300ko
fixes #283 

Capture d’écran 2022-08-17 à 12 58 56

ciphrd:
LGTM
Great and much needed refacto, thanks :)

PR #345, opened 2022-08-28
feat(article): add mentions to article

### LINKED ISSUE #342: [Articles] Mentioning users with `@tezosaddress`
The purpose of this feature is to add tezos address mentions as a leaf block type.

# Overview

## Tezos address format

A tezos address is a 36 characters base 58 encoded string.
The following characters are used for the base 58: `123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz`
A tezos address can only contain these characters, and starts with `tz1`, `tz2` or `tz3`

## A valid tezos address mention

* A mention a `@` character directly followed by a tezos address.
* A tezos address mention can only be set in a paragraph.

## Auto-completion

When a user writes a `@` character followed by 3 characters, it opens an autocompletion dialog, allowing them to pick a user like on github:

image

When a user in the list is selected, its address is inserted in the markdown after the `@`

There's already a component we can leverage for that:

image

## Display

Instead of displaying `@tz1...AB`, we display the user name linking to their profile. This is trivial for the viewer, but a bit more complex for the editor. When a tezos mention leaf is inserted, the editor displays an inline block which is completely deleted if the user presses the backspace at the end of the user name.

# Implementation
- [x] add parsing of valid tezos address mentions into leaf blocks for viewing articles
- [x] tezos mention component to fetch a user based on the address to render its username
- [x] auto-completion of a mention when typing `@`, as described above
- [x] tezos address mention leaf is deleted when pressing backspace after it

GemN:
fix #342 

It was tough to implement (lot of processors trial and errors to make it works everywhere)
Double check for the style of mentions / floating menu


ciphrd (changes requested):
Nice implementation, very functional.
I made a few graphical changes so that it's more readable (replaced UserBadges by some links):

image

---

@GemN I noticed one issue though, if there is a space character after the mention, it disappears in the viewer (but not in the editor). See:

Editor, good:
image

Viewer, space not rendered:
image

GemN:
Should be good now - rehypeFormat was breaking (removing whitespace when it should not)

ciphrd (approved):
LGTM

PR #347, opened 2022-08-30
allow js modules in the sandbox

### LINKED ISSUE #246: [sandbox] Missing/wrong MIME type breaks JS/ESM module scripts
Already reported this issue a while ago via DM to @ciphrd, but the sandbox doesn't seem to support JS/ESM modules anymore and after further digging it could have to do with the fact that the MIME type for these files is missing/misconfigured on your server/CDN and Chrome reports the following error:

> Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.

Btw. Testing the project via the the minting UI workflow (uploaded to IPFS) works fine and the JS file is served with the proper headers...

Solution: Ensure this HTTP header is defined for all JS files served: `Content-Type: text/javascript`

maerzhase:
- fix #246

set proper mimetype for js files and fetch module files if referrer differs but file is part of cache. 

If you don't have a js module project you can use this zip to test it:
[Archive.zip](https://github.com/fxhash/fxhash-website/files/9455112/Archive.zip)

### DIFF
--- a/public/sandbox/worker.js
+++ b/public/sandbox/worker.js
@@ -1,6 +1,13 @@
 const cache = {}
 const referrers = {}
 
+async function fetchUrl(url, file) {
+  // Let the service worker decide on secure response
+  // headers, but set its body to the file blob.
+  const record = await fetch(url)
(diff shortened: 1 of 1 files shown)


ciphrd (approved):
LGTM

PR #359, opened 2022-09-01
fix(article): remove entire nodes when deleting foward/backward for medias (figure) and table

### LINKED ISSUE #354: [Articles] Application error when editing an article, edge case
* base branch: `dev`

I'm forwarding a bug reported by oleg on discord, I did not try to replicate it:

> if you hit backspace twice in an empty paraghaph and there will be an image without caption before it, the page will broke with application error.
> 
> but this time it killed all my draft body. All images (already loaded into IPFS) lost, all text too. This happen when i was editing it, means after checking preview.

GemN:
fix #354


ciphrd (approved):
LGTM

PR #361, opened 2022-09-01
fix(article): add responsive for tezos-storage-pointer

### LINKED ISSUE #356: [Articles] Bad responsive for Generative Token & Gentk embed tezos-storage-pointers
* base branch: `dev`

image

- [ ] on mobile, hide the artist avatar
- [ ] make sure that the whole artist name goes back to the line if doesn't fit on the width

GemN:
fix #356


ciphrd (commented):
@GemN I don't understand the necessity of preventing submitting the form on `ENTER`, why did you add this ?

GemN:
If your article is valid (meaning everything required is already filled), if you navigate in content body and use embed media, tezos-pointer-storage (the research) or anything with an input and press enter, it will trigger the form submit button.

Capture d’écran 2022-09-06 à 10 16 03

If you press enter while filling tags or a research by mistake I don't think the expected behavior is to trigger the preview button and change page

ciphrd:
@GemN 

Maybe this should be solved in a more robust way, by stopping the propagation of the events whenever a form submit is detected.
This way we don't hack the default browser behavior and we make sure every component susceptible of triggering a form submit stays in its own scope.

ciphrd (changes requested):
See my previous comment

PR #365, opened 2022-09-02
fix(article): invalid slug with just a dot redirect to id

### LINKED ISSUE #364: [Articles] Some slugs cannot be set in the url
* base branch: `dev`

Can be tested on article id 321

The article title is `𝗔𝗡𝗔𝗩𝗘𝗥.𝗦𝗘` which was sluggified to `.` because of unsupported characters. However, the "." cannot be used next to a "/" in the url as it is removed by the browser:

`https://fxhash.xyz/article/./` is turned into `https://fxhash.xyz/article`

# Resolution

* detect edge cases where all the sug characters cannot be added to the url
* use the ID url instead

Function to update: `getArticleUrl`

GemN:
fix #364

### DIFF
--- a/src/utils/entities/articles.ts
+++ b/src/utils/entities/articles.ts
@@ -8,11 +8,15 @@ const flagged: ArticleFlag[] = [
 ]
 
 export function getArticleUrl(article: NFTArticleInfos): string {
-  return article.slug && !isArticleFlagged(article)
+  return isValidSlug(article.slug) && !isArticleFlagged(article)
     ? `/article/${article.slug}`
     : `/article/id/${article.id}`
 }
 
 export function isArticleFlagged(article: NFTArticleInfos): boolean {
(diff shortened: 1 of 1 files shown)


src/utils/entities/articles.ts
```diff
@@ -8,11 +8,15 @@ const flagged: ArticleFlag[] = [
...
     : `/article/id/${article.id}`
 }
 
 export function isArticleFlagged(article: NFTArticleInfos): boolean {
   return flagged.includes(article.flag)
-}
\ No newline at end of file
+}
+
+export function isValidSlug(slug: string): boolean {
```
ciphrd, line 20: The rule can be extended to "is valid if it doesn't contain a slug".
There are a few other iterations of slugs not being valid when they contain a dot.

ciphrd (changes requested):

PR #368, opened 2022-09-05
feat(article): add markdown editor and markdown editor with preview

### LINKED ISSUE #180: WYSIWYG editor + viewer for extended Github Flavored Markdown
## Abstract

We are soon going to introduce articles as NFT on the platform, and so we would need a state-of-the-art editor to edit Github Flavored Markdown (GFM). We will develop this feature by iterating progressively on it, as the direction may shift during the dev. We will need to find great modular libraries we can easily extend and fully customize.

This feature will be a core component of the fx ecosystem, and as such we need to implement in a clean, robust, stable and ux-friendly way. People expect their web-based text editors to be state-of-the-art.

## The editor

We want the editor to behave similarly to Notion text editor, supporting markdown, keyboard shortcuts and the like. We want a WYSIWYG editor, producing GFM under the hood. Notion is great because it provides native support for markdown as well as UI tools to manipulate the content.

## Regular medias

We also want to provide the ability to embed image, audio and video content. Files will be stored in-memory during the redaction of the article, all the medias will be uploaded at once when publishing the article (for later). Users should have the ability to drop medias in the article to add them.

## Blockchain-media spec

We propose to extend the markdown specification to have a syntax specifically dedicated to the referencing of blockchain smart contract storage data: the blockchain-media spec We plan to use this to display fxhash tokens in the body of the articles (the spec is also designed to provide inter-operability). The following document describes this new syntax: https://www.notion.so/fxhash/Article-content-spec-5fe1029e69b748b1beac36db3b3f20d7

The viewer should be designed to parse the markdown and generate a JSON object from the blockchain-media spec:

IN: 

```markdown
{address:KT1RJ6PbjHpwc3M5rw5s2Nbmefwbuwbdxton;key:741172}(hello text)
```

OUT:

```json
{
  "address": "KT1RJ6PbjHpwc3M5rw5s2Nbmefwbuwbdxton",
  "key": "741172",
  "annotation": "hello text"
}
```

For the first iteration, the viewer will only need to detect a valid blockchain-media string, turn it into a node and output a json object rendered by a custom react component. In the next iteration we will build viewers from the json data.

## Implementation

We will be working on a separate `feature/markdown-editor-viewer` as we'll keep iterating on the feature. A page `/editor` will be created to ease the development. We need to discuss on how to extend existing markdown parsers so that we can parse the blockchain-media spec, and ideally use react components to render these blocks. AFAIK there are lots of markdown parsers but they output html, we would like the ability to somehow have react components be used for rendering in there (because the Blockchain-media blocks will need network calls to be rendered).

### LINKED ISSUE #327: [Articles] Is it possible to have a (optional) markdown editor instead of the WYSIWYG for power users?

GemN:
fix #327


ciphrd (changes requested):
Great first iteration, looks good.

I'll ping lomz for some design with the tabs (maybe top left ?)

I'll leave some remarks regarding some of my findings when testing the feature.

# P inside blockquotes

I believe this is the expected behavior:
image
image

The blockquote is parsed with a p inside of it.

**Solution:**
* remove margin on first/last elements of a blockquote
* hide tooltip if paragraph inside of a blockquote

*Maybe this should be fixed at once with the general issue  about blockquotes and how we handle them*


# Format destruction

If i input this into the markdown editor:

```md
# Trigger

The trigger defines when the capture module will take the preview after loading the token in a web browser:

* **Fixed delay**: Give it a delay of X seconds, and once the project is loaded, the capture module will wait X seconds before triggering the capture
* **fxpreview()**: The capture module will wait until your code calls `fxpreview()`. As soon as the function is called, the capture will be triggered. You can call this function whenever your algorithm is ready to be captured. *The capture module will automatically take a capture after 300 seconds have passed after your project was loaded in the browser.*
```
image

Then, I switch to the WYSIWYG editor, I focus any paragraph, and get back to the markdown editor, the content is butchered:

image

Which actually break the output completely:

image

Repeating the manipulation will reformat the markdown until a stable state is reached, but markdown is destroyed.

This is a big issue, probably far from being easy to solve. The problem is that slate takes over the markdown and its format. We would need a robust solution where the markdown being inputted is not destroyed when doing some edits with the WYSIWYG editor.


# Loss of history context

We lose the WYSIWYG history context when switching from the WYSIWYG editor to the markdown one


# Lack of padding at the bottom of the preview on the side-by-side view

It would be more comfortable to have some padding there.

image

GemN:
About the loss of history, when we switch we recreate an instance of the editor parsing the markdown edited. 
Pros: 
- The wysiwyg is recreated from a markdown, so it behaves like a first load (trigger on init, normalisation etc)
Cons: 
- No history

A possibility is to keep the instance in memory, hide the editor and update the core value. 
Should I go with that?

PR #372, opened 2022-09-06
fix(article): update media size (video, image)

GemN:
clarity


GemN:
fix  #386 (height dimensions of medias)

PR #398, opened 2022-09-15
fix: optimize infinite scroll - doesn't trigger empty queries when under 20 results

### LINKED ISSUE #389: Optimize & refacto Infinite Scroll
* base branch: `dev`

This issue aims at solving 2 problems, one being functional and the other being refacto.

# Overview

https://user-images.githubusercontent.com/18054507/189650153-9b7bd5b3-20c6-4888-8bf0-9443aaf5151b.mp4

# Too many fetches from infinite scroll

## Problem

https://user-images.githubusercontent.com/18054507/189644390-b556a500-aa0c-43c3-9e5f-1f817934b3f1.mp4

Infinite scroll will trigger one time more than what's needed in most of the cases. Regardless of the response from the backend, it will run another query and prevent loading more only if there is an empty response. 

## Solution

If the response from the backend is under the pagination `take` parameter, we already know that we won't receive more from the backend, and as such we can set terminate the InfiniteScroll loading process.
**Drawback**: Let's say there are only 20 items we can get from the back end, and we set the `take` parameter to 20, there is no way for the frontend to know that we've reached a limit. This would require a `count` parameter on the backend but there is none currently so we have to accept that in some circumstances there will be an empty trip to the backend. For most of the cases, this won't happen.

## Careful !

Due to how we handle pagination, there may be some shifts in the results and we cannot rely on the **results** of the cache merge to count how many items were fetched from the back. Let me explain with a schema:

Let's say we want to fetch the generative tokens:

```
DATA IN THE DB, ordered by id DESC

300
299
298
297
296
295
294
293
292
291
290

If we run a query with 
SKIP: 0
TAKE: 3

300  <-- start query 1
299
298  <-- end query 1
297
296
295
294
293
292
291
290

If the data on the backend stays the same, when we do the next fetch:
SKIP: 3
TAKE: 3

300
299
298  <-- end query 1
297  <-- start query 2
296
295  <-- end query 2
294
293
292
291
290

However, if there is a change of the data in the DB and a new item is inserted, there will be a shift in the pagination:
SKIP: 6
TAKE: 3

301  <-- ! new item was added !
300
299
298
297
296
295  <-- end query 2   <-- start query 3
294
293  <-- end query 3
292
291
290

This will result in the front receiving the item "295" twice
```

This is why a cache merge policy is in place to remove duplicate from being merged in the results:

https://github.com/fxhash/fxhash-website/blob/05b2d4c7f8aae85e3a9f778107068906c3bd9bea/src/services/ApolloClient.ts#L17

**As such, we cannot rely on counting the number of items added to the cache every time it changes to optimise the new refetch policy.**

# Refacto of the InfiniteScroll policy

## Problem 

Some code is duplicated throughout the app to handle the refetch policy, with very minor changes (see first video).

## Solution

We can abstract this code (maybe eventually put it in a hook or in the InfiniteScrollTrigger component).
I believe it makes sense to implement everything in the same PR as both issues can be solved at once. However, it needs to be carefully designed to make sure that we don't break any fetch mechanism.

GemN:
fix #389


ciphrd (approved):
LGTM
I will keep the issue opened until the refacto is done, but I will merge this branch.

PR #406, opened 2022-09-24
refactor apollo client singleton to new instance

### LINKED ISSUE #396: Always use new apollo client on server
To ensure fetching latest data on the server we must create a new apollo client for each server request. This might explain a few weird caching situations that where reported by users, e.g. articles not returning the latest metadata etc.

Currently same apollo instance is reused on the server.
https://github.com/fxhash/fxhash-website/blob/f80bfe077564a38cbf70b8b8d5e90503e7109fd0/src/services/ApolloClient.ts#L3-L8

maerzhase:
on server a new client must be created to ensure serving latest data to users.

- fix #396


ciphrd (approved):
LGTM!

PR #410, opened 2022-09-26
fix(article): sanitize urls

### LINKED ISSUE #409: Fix XSS on embed-media element
@GemN 

See this: https://www.fxhash.xyz/article/id/531

The XSS is blocked by chrome but is successfully injected into the DOM, so there may be a vector that needs to be sorted out.

It's using the embed element component in a weird way.

```md
::embed-media[ClickMe]{href="javascript:alert(1)"}
```

Source: https://gateway.fxhash2.xyz/ipfs/QmWC9i1KP3LnLSMbvz9jmA8hdNY86q9UpZ1tgMkZP8GZSg

These URLs should be sanitized before getting injected into the DOM, see these helpers: https://github.com/fxhash/fxhash-website/blob/main/src/utils/url.ts

GemN:
fix #409 

WIP, need to check if it's happening on other custom components


ciphrd (changes requested):
The sanitize breaks some components, for instance the images are not working properly I believe, see this one for instance:
http://localhost:3000/article/e38-system-of-a-down-market

image

ciphrd (approved):
LGTM

PR #427, opened 2022-10-07
Enhancement/optimise images

### LINKED ISSUE #408: Optimize the service of images for thumbnails & display a blurred placeholder while loading
* base branch: `dev`

As discussed internally, we want to have blurred placeholders & optimized image fetching based on available screen space. The infra to support this features **has only been deployed in the dev environment**.

This should be implemented whenever a Card is displayed:
- generative token card
- objkt card
- article card (small & big)
- user badge

# Image placeholder

The API now exposes a new Media Image field for every entity with which an image is associated:
- article: `thumbnailMedia`
- generative token: `captureMedia`
- objkt: `captureMedia`
- user: `avatarMedia`

All of these fields point to a Media Image abstraction, which exposes information about the media such as the resolution, base64 placeholder & mime type. It should be noted that this entity is processed in parallel of the indexing, as such sometimes this data may not be available. The implementation should support that. There are 3 cases, based on the state of the `processed` field of the entity:
- if **false**, media has not yet been processed
- if **true** and **width != null**, media has been processed and data is available
- if **true** and **width is null**, media processing has failed

The placeholder field is a base64 string of a 12x12 downscale of the initial resource.

# Image optimization

Instead of fetching the images from the main gateway all the time (which is quite expensive especially when navigating the iterations of a project), we will now fetch images from our serverless media processing endpoint. **Images should be fetched based on the screen space they will take**.

Our media endpoint is the following (in dev): `https://media.dev.fxhash-dev.xyz/{transforms}/{cid}`
* transforms: a list of transforms separated with commas
* cid: the ipds cid (not the whole uri just the Qmzefzef...

For now we can only resize through this endpoint, it can be done with the following filter: `w_500` for 500px. It will resize the image to 500px of width while preserving its aspect ratio.

## Caching optimization

Processing images on-the-fly is fast, but to maximize efficiency and reduce our costs we would like to avoid this processing as much as possible. We have a cache on this media endpoint, and this cache will only hit if urls are the same of course. To ensure that we hit this cache as often as possible, I propose rounding the image size requested to the 50th closest step. Ex:
- need to display 310 px -> request 350px
- 401px -> 450px
- 587px -> 600px
I believe this is quite optimized strategy, and will ensure we always serve high quality images to the user's display.

# Design decisions

This features should be well designed as it's going to be a major upgrade for us, and will improve UX by a lot. Should we create our own image component to implement this custom strategy ? Should we use next/image only (I think it's covered by a [custom loader](https://nextjs.org/docs/basic-features/image-optimization#loaders) ? What about the effect, is their one to make a smooth transition between the placeholder and the actual image ? Or is it too expensive ?

# TODO summary

- [x] display blurred placeholder when loading an image
- [x] handle cases where placeholder/width/height are not available
- [x] fetch the medias from the media.fxhash endpoint instead of gateway, with appropriate size
- [x] un-blur smooth animation
- [x] apply new image strategy on the cards: generative token, objkt, article
- [x] apply strategy on: user badges

maerzhase:
- fix #408 

This PR brings the image optimisation to fxhash. 

@ciphrd as mentioned i've refactored the queries a little with more fragments (especially for generative tokens). I also moved all the queries that were directly declared in components.

For the whole image optimisation feature i've added a `Image` component which we are now using ~90% across the application. It has a build-in fallback to lazy load the image directly from ipfs if media information are not present. It uses `next/image` internally which solves many of our problems. Next image creates a srcset and fetches images based on the sizes we provide in the `next.config`. This solves our problem with caching because resources are only requested with this subset of sizes. 

The component has handlers to receive error and onload which makes it easy to implement some custom loading effect, but I didn't add a loading effect directly to the component. 

Do we already have an idea how we would handle images in articles? I wasn't sure about that. 

One idea for improving the image api: If we make it possible to provide only one dimension and return the image i correct aspect ratio it would make it easier to use the image api when width + height information are not (yet) present. 

Kapture 2022-10-07 at 18 12 33


ciphrd:
Closing this as #428 fork has been merged.
Thanks for handling most of the implementation

PR #435, opened 2022-10-14
fix: image component get border-radius inherit

GemN:
fix bug related to loading square flashing in image components


ciphrd (approved):
LGTM

Changed the "force" strategy implement which would bypass the actual image resize API in favor of storing resolution per CID.

PR #436, opened 2022-10-14
fix(article): parsing list item interpret bold, emphasis and directive as expected

### LINKED ISSUE #429: [Articles] Special formatting in article component List break List on reload
Adding the following, will break list on reload (maybe due to the parser or normalisation):
```md
- **test**: test
- test
```

Capture d’écran 2022-10-10 à 10 02 13
Capture d’écran 2022-10-10 à 10 02 20

GemN:
fix #429 
First iteration to solve the list bug

I explored many paths:
- Add a custom mdhast plugin that add a paragraph as a list item children containing his children: didn't work (directive were still parsed instead of ignored like in a _paragraph_)
- Remove flatten paragraph plugins `mdastFlattenListItemParagraphs` and work around **slate paragraph** element: **slate paragraph** have already his own set of rules and too complicated to interact (need to check parent and childs always leading to many new edge cases)
- Keep flatten listItems, add back a paragraph in the `slateToRemark` plugin: **this work** but the solution feels weak, what will happens if we use the markdown parser in the future and add nested lists, weird leafs etc. However that solves every other bugs since now it's treated as a paragraph (like in the original parser)

I need to a find a way to add add a paragraph in between and treat childs as is instead of parsing myself if anyone has a more robust idea: (@maerzhase ?)

```ts
export const listItemProcessor: IArticleElementProcessor = {
  transformSlateToMarkdownMdhast: (node: any) => {
    return {
      type: "listItem",
      spread: node.spread,
      children: [
        {
          type: "paragraph",
          // I don't want to parse manually, I'd rather call a function that will use existing rules
          children: node.children.map((child: any) => {
            const text = {
              type: "text",
              value: child.text,
            }
            if (child.strong) return { type: "strong", children: [text] }
            if (child.emphasis) return { type: "emphasis", children: [text] }
            return text
          }),
        },
      ],
    }
  },
}
```


ciphrd (approved):
LGTM

PR #448, opened 2022-10-27
fix(filters): iterations show filters in mobile

GemN:


maerzhase (approved):
In general looks good to me! 👍 🐰 

Minor remark: The filter overlay has the same problem like the main menu—when the modal is open the body is still scrollable. I opened a separate issue for that

- #462

PR #449, opened 2022-10-28
Fix missing image media entities

### LINKED ISSUE #441: [BUG] Some entities are not loaded through their image media element
* base branch: `dev`

Some entities are probably missing their media element from the API query, resulting in places where we still load the full size image. I compiled a small list but there might be other places in the application, a full pass may be required to check everything.

### Owner avatar on the gentk page

/gentk/[id]

image

### Project thumbnail in the incoming section

/explore/incoming

image

### Gentk thumbnails in the reveal feed

/explore/reveal-feed

image

### Gentk thumbnail & owner avatar in the project marketplace page

/marketplace/generative/[id]

image

### Gentk thumbnail in the offers tab of the project marketplace page

image

### Same page, activity feed, user avatar

image

maerzhase:
- fix #441 

@ciphrd made another pass on the whole app. i fixed all placed you mentioned in the issue. i refactored all queries define in components into the `queries` folder and use the right fragments there. I also added the `<Image />` component to the `TezosStorage` element in the articles. 

### Questions

I am not sure if you already thought about it, but how shall we handle images in the articles? Can we do that at all?

Should I also refactor the query from the `LiveMintingEvent` container? Do we actually have `captureMedia` on them?

https://github.com/fxhash/fxhash-website/blob/cdb5153c6a4f727db4def5e480a3a171c2956929/src/containers/LiveMinting/LiveMintingEvent.tsx#L14-L40

### DIFF
--- a/src/components/Image/index.tsx
+++ b/src/components/Image/index.tsx
@@ -15,6 +15,8 @@ import {
 } from "../../services/Ipfs"
 import { MediaImage } from "../../types/entities/MediaImage"
 
+export const OG_IMAGE_SIZE = 1200
+
 // a list of common sizes which will be used to fetch the resource, ensuring
 // we hit the cache as often as possible
 const sizes = [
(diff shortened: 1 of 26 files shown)


ciphrd (changes requested):
LGTM

There is still the og tags in the <meta> tags where the olg gateway is loaded I believe, this can be changed to optimize for SEO :)

ciphrd (approved):
LGTM

PR #454, opened 2022-10-28
decode search query string

### LINKED ISSUE #453: Search query not decoded
The search query string seems to be not decoded when returned from getServerSideProps

maerzhase:
- fix #453


GemN:
LGTM

PR #456, opened 2022-10-30
fix error on users query due to wrong params

maerzhase:
@ciphrd wrong query params resultet in user query never being resolved and therefore the `royaltiesSplits` was always empty preventing the possibility to mint articles.  

mentioned in discord by ErikWestener:

> when possible can someone check errors on fxtext. A few folks are unable to mint new pieces, myself included.  Attempting on multiple browsers, clearing cache, etc. -- I know it's not much to go on but error is below. 
image


GemN (approved):
good catch, sounds like the skip/take has been badly merged (or added by myself)

PR #460, opened 2022-10-31
fix: dark mode is applied on body instead of nested div, which fix theme issues within react portal

### LINKED ISSUE #455: Mobile navigation does not implement dark-mode preference
Having darkmode turned on, opening navigation results in light-mode nav component.

GemN:
fix #455


maerzhase (approved):
LGTM! 👍 🐰

PR #461, opened 2022-11-01
fix(mobile): add support for iOS mobile, search opens and focus correctly, input don't focus anymore

GemN:


maerzhase (approved):
Nice good job! 👍 🐰

PR #464, opened 2022-11-03
fix(image): pixel density used is maximum 2, maximum thumbnail size used is 1400px

### LINKED ISSUE #446: [Images] Limit pixel density for image api
Cam up from a discussion in discord.

As image API endpoint has problems with too large images we should limit the pixel density to 2x on the frontend.

GemN:
fix #446


ciphrd (approved):
LGTM

PR #465, opened 2022-11-03
Change Header style to fix issue#462

as00r-dev:
Right now, header is not fixed to top on desktop view, and there is a gap between header and menu in mobile view.

To fix this, i've fixed the header to top and added some margin to main to make this work.


as00r-dev:
Deleted this, because accidentally committed changes made by prettier to format the code which caused 9k+ lines of change on package-lock.json

Issue #462, opened 2022-11-02
Bug: Prevent body scrolling when modal is open


maerzhase:
When a modal opens (e.g. mobile main menu) the body is still scrollable which is irritating. 

![Kapture 2022-11-02 at 13 57 07](https://user-images.githubusercontent.com/1128485/199495482-1db7147d-cc1c-42fe-a7f8-2802936d62fb.gif)

PR #466, opened 2022-11-03
Bug/issue#462

as00r-dev:
Right now, the header is not fixed on top, this causes the mobile
version to glitch which can be viewed in Issue https://github.com/fxhash/fxhash-website/issues/462.

To fix this, i have made header fixed on top and added a margin of
140px to the main element to accommodate for the fixed header.

### DIFF
--- a/src/components/Layout/Header.module.scss
+++ b/src/components/Layout/Header.module.scss
@@ -7,6 +7,11 @@
   min-height: 140px;
   padding: 18px var(--main-padding-sm);
   padding-left: 10px;
+  position: fixed;
+  top: 0;
+  width: 100%;
+  background: white;
+  z-index: 999;
 
   @media (max-width: $breakpoint-md) {
     min-height: 90px;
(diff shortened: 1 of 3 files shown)


maerzhase (commented):
@as00r-dev first of all thanks for participation.

I slightly disagree that the PR solved the linked issue, because the body will still scroll also if the header is sticky. We need to prevent scrolling the body when the body or more general modals are opened.

in general sticky header is a design decision i'd like to get some input from the team. @fxhash/product what do you think

as00r-dev:
@maerzhase Hi. Thanks for the feedback. 

We can toggle `body.style.overflow` to the `body` element when user clicks on Hamburger. 

```
const MENU_MODAL  = document.querySelector(".MobileMenu_container__XVSEy");
const BODY = document.querySelector("body");
const MENU_BTN = document.querySelector(".Navigation_hamburger__58XZY")

MENU_BTN.addEventListener("click", (e) => {
  if(MENU_MODAL.classList.contains("MobileMenu_container--open__fg3N1")) {
    BODY.style.overflow = "hidden";
  } else {
    BODY.style.overflow = "auto"
  }
})
```
I wanted to add this. But, i was not able to find the file where this click is being handled.

as00r-dev:
I am closing this PR for now.

PR #473, opened 2022-11-05
bug/ _ Fix darkmode dark letters

### LINKED ISSUE #470: Regression: When in dark mode, some text that should be light still appears as dark
- Go to fxhash website
- Open settings
- Turn on dark mode.

Notice that some text in dialogues and modals appears as black on a dark background, making it unreadable.

image
image

daniel-bergmann:
This change resolves #470 by adding some css rules to globals.scss.

### DIFF
--- a/src/containers/Settings/Settings.module.scss
+++ b/src/containers/Settings/Settings.module.scss
@@ -10,7 +10,7 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-
+  
   & > *:first-child {
     width: 60%;
   }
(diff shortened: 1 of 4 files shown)


GemN:
Hi @daniel-bergmann,

Thank you for your time and making a PR, however the problem on this case was that the text color wasn't applied our modals (and other things) that use ReactDOM.createPortal. Setting the css variable on body will correctly apply the default text color on current and futur Portals.

This is fixed now by #475

PR #475, opened 2022-11-06
fix: modal and portal get correct font-color

GemN:
Modals and anything using portal wouldn't have their default color set to --color-black, meaning that they would still be default browser color when using dark-mode. This PR fix that


maerzhase (approved):
looking good 👍 🐰

PR #485, opened 2022-11-16
feat: marketplace pagination, arrows on desktop, touch on mobile

### LINKED ISSUE #484: [Marketplace] Increase the number of projects available in the "highest volume" component
Display more project using a pagination with arrows on desktop, swipe gesture on mobile

GemN:
fix #484

### DIFF
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -50,6 +50,10 @@
 
 .font-regular {}
 
+.size-custom {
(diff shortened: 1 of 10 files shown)


maerzhase (approved):
Good job! @GemN 👍 🐰 

from specs we should create two new issues:
- Increase limit on marketplace api to X (e.g. 100) to load more than 50 
- Add sort by volume on the collections tab 

Both issues need work on the api.

PR #493, opened 2022-11-21
Fix/#487

### LINKED ISSUE #487: [Bug] App throws error on accessing type of undefined
Pointed out by louis on discrod:

# View of error

# Stacktrace

maerzhase:
- fix #487

I cannot reproduce the problem. But from the error description we got this should prevent it from happening...


GemN:
I don't know if it's a good fix - this fix the symptoms for sure but at some points the type should be an user, and not null. The check may be needed upper in the parent

GemN (approved):

PR #504, opened 2022-12-02
replace cover with useClickOutside hook

### LINKED ISSUE #503: Moderation Dialog: Scrolling on modal doesn't work
reported by @liamegan 

> Moderation dialogue is broken, when opening the reason a scrollbar appears but you can't scroll.

maerzhase:
- fix #503 

We are using a `<Cover />` component when select is open that handles closing the open select dialog when you click anywhere. Since the `<Cover />` lays ontop of the scrollable area the scroll event doesn't arrive in the scrollable container. 

I removed the `Cover` from the `Select` component and replaced it with the `clickOutsideHook`

### DIFF
--- a/src/components/Input/Select.tsx
+++ b/src/components/Input/Select.tsx
@@ -45,7 +45,6 @@ export function Select({
   ...props
 }: Props) {
   const selectRef = useRef<HTMLDivElement>(null)
-  const buttonRef = useRef<HTMLDivElement>(null)
   const [searchString, setSearchString] = useState<string>("")
   const searcherRef = useRef<FuzzySearchType<any> | null>(null)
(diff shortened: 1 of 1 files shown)


src/components/Input/Select.tsx
```diff
@@ -115,6 +115,8 @@ export function Select({
   // what are the options displayed ?
   const displayOptions = searchResults || options
 
+  useClickOutside(selectRef, () => setOpened(false), false)
```
GemN, line 118: Skip "useClickOutside" when options are not displayed

GemN (changes requested):

GemN (approved):

PR #509, opened 2022-12-07
Feature/press kit page

maerzhase:
Implements the press-kit page based on figma designs and content provided by team

### Url schema

- `/press-kit` -> fxhash
- `/press-kit/brand` -> brand
- `/press-kit/logo` -> logo
- `/press-kit/media` -> media kit

### New tabs layout

@GemN @ciphrd 
I added a new layout to the tabs called `fixed-size-narrow`.

Screenshot 2022-12-07 at 21 28 04


GemN (changes requested):
## Desktop 

Pressing "back button" on desktop should move back to the previous section (it only update the url at the moment)

Capture d’écran 2022-12-14 à 16 55 02

- Errors on some videos 
- The display could be a bit spaced in case of errors

Good work on desktop! 

## Improvements

I think every event / exhibs / museums should be dated and ordered by most recents.
I will make a copy in discord of this since it's related to product
Capture d’écran 2022-12-14 à 17 12 03

## Mobile
I have a few suggestions for mobile since it has not been provided in figma

### logo

Capture d’écran 2022-12-14 à 17 04 51

- "full logo" and other H2 should be 20px on mobile (refer to the mobile version on figma, try to match titles size, ping on discord if needed)
- text-align center for download buttons with a bit of padding (check with viewport width 320px)
- "Download the full presskit at once." download .zip should be block width 100% under

### medias

Capture d’écran 2022-12-14 à 17 19 13

- like in logo, update h2 titles as expected
- when an dropdown is open, let the dropdown be position: sticky on top, that way we always have the title of event and we can easily close it by taping on it

Capture d’écran 2022-12-14 à 17 20 56

- Images should be clickable and open in a new page (or as a modal preview with cover if you feel like it)

maerzhase:
@GemN thanks for your helpful review. i did the mobile pass and added all the other things.
instead of giving more space for the errors on video load i decided to hide them. 
back/forward navigation should work now.

GemN (changes requested):
Capture d’écran 2022-12-15 à 13 30 57
Capture d’écran 2022-12-15 à 13 30 50

There is a bit of extra gap in mobile tab nav, check on the explore tab, underline must finish where the next item begin, and there is no extra padding on the last element.

The sticky dropdown works well in desktop however in mobile, it has some weird and inconsistent behavior (check video)

[screen-recording.webm](https://user-images.githubusercontent.com/13420650/207869065-ea7c2f54-2781-40ed-8b0b-4d0cf5aacdff.webm)


As an improvement, when a dropdown collapse scroll back to the origin position of the dropdown, else we lose the focus in the page because of the layout shift

GemN (approved):

GemN (approved):

PR #513, opened 2022-12-13
feat: add url /my/... that redirect on profile and their subroutes

GemN:


maerzhase (approved):
✅  tested locally and on preview 
looking good 🐰 !

PR #518, opened 2022-12-20
fix(article): bad url are caught instead of throwing the page

GemN:


maerzhase (approved):
jeah thats the right way of doing it. sorry i should have spotted this before. 
we always need to be careful at the spots where we depend on the user input. it can have any kind of unexpected behaviours. 

👍 🐰

PR #523, opened 2022-12-26
feat: add a password protection page for staging/preview branches

### LINKED ISSUE #522: Custom password protection
* base branch: `dev`

# Current solution

Right now, we are using vercel's password protection feature, which is not very good:
* a single password for all the preview branches
* 150$ / mo

image

While it was good, we will now need a more granular approach to secure our preview branches.

# Proposed solution

* Implement a site-wide protection if the following environment variable is defined: `PASSWORD_PROTECTION`. The value of the env variable will be the password required to authenticate.
* The session is preserved with an http-only cookie
* If we change the password, the connection should be invalidated (we can derive a secret string from the password so that when the password is changed, it changes the secret key used to create and parse the user cookie)
* Focus on speed: please find a solution where the least amount of code is executed if PASSWORD_PROTECTION is empty or undefined. Maybe there is a way to set this in the config, or only bundle some code at build time if the env is defined. The takeaway is that the prod environment should not be impacted at all by this.

GemN:
Feature: password protection page for deploy / staging / dev branches.
Use env variable PASSWORD_PROTECTION=mypassword to set.

- middleware based (happens on all routes matched in the config, even if no SSR)
- fast when unset
- cookie based
- store password as jwt in the cookie
- dynamic import of App.js when password-protection is on (quick display of protection page, with no dependency except global css)
- robust against node timing attack

Next has been upgraded to 12.2.0 with no visible breaking changes

fix #522 

Conditional matcher based on env variable doesn't seems to work? Tried to pass null / array of unknown routes, middleware would still trigger on all pages.


maerzhase (approved):
Great work @GemN works like a charm. I've merged dev and fixed the build issues.
@ciphrd you have to remove vercel password protection and setup branch based envrionment variables: https://vercel.com/changelog/environments-variables-per-git-branch

ciphrd:
LGTM!
Let me know when it's merged and I'll add the envs

ciphrd (approved):

PR #533, opened 2023-01-09
feat: add donation address for cure parkinson's

GemN:
Add new donation address as per discord request


maerzhase (approved):
good good 👍 🐰

PR #552, opened 2023-01-24
feat(user): add info message when a gallery / container is empty

### LINKED ISSUE #115: [UX] Add an empty message when there is no creations in user profile
When the user has no creations, there is just a blank space. I think we should display at least an empty message (ex: "'username' has not made any creations yet, come back later!").

Currently can be confusing to see nothing (can make you think of an error while loading data)

Reference: https://www.fxhash.xyz/u/NukCZ

Capture d’écran 2022-05-05 à 18 18 18

GemN:
fix #115


maerzhase (approved):
ty looks good.

just for reference thats how it looks like:
Screenshot 2023-01-27 at 11 09 16

PR #553, opened 2023-01-24
feat: add 'mint article' in the profile nav'

GemN:
as agreed with qaulv on discord, adding a shortcut to the profile dropdown to emphase on the article creation


maerzhase (approved):
👍 🐰

PR #556, opened 2023-01-26
feat: add winter checkout to secondary sales

GemN:
waiting on small fix from winter team to get the wallet address to redirect to after the purchase


maerzhase (approved):
lgtm! tested on testnet with test card, works! 
i've removed a console.log @GemN

PR #557, opened 2023-01-27
Feature/save preferred sorting setting on marketplace generative

### LINKED ISSUE #554: Save preferred sorting setting on marketplace generative
mentioned by liquid:
> Unrelated, but is it possible to save the filtering used in a project marketplace page during a reload? 
> It always defaults to recently listed but that's rarely useful 😦 

In further discussion we figured we are referring to the sorting settings. 

Screenshot 2023-01-25 at 19 41 34

We should store the setting in  localstorage and also put it in the query params to allow url sharing

https://www.fxhash.xyz/marketplace/generative/24169

Jobayer12:
The PR fixes #554 
The save preferred sorting setting on marketplace generative has been implemented.

### DIFF
--- a/src/containers/Marketplace/GenerativeListings.tsx
+++ b/src/containers/Marketplace/GenerativeListings.tsx
@@ -44,15 +44,18 @@ function sortValueToSortVariable(val: string) {
     [split[0]]: split[1].toUpperCase(),
   }
 }
+interface Query {
+  sort: string
+}
 
 interface Props {
   token: GenerativeToken
-  urlQuery: Record<string, string>
+  query: Query
 }
 
 const localStorageKey = "marketplace_generative_sort"
-const getSortFromUrlQuery = (urlQuery: Record<string, string>) => {
-  const { sort } = urlQuery
+const getSortFromUrlQuery = (query: Query) => {
+  const { sort } = query
 
   // if there is a sort value in the url, pre-select it in the sort input
   // else, select the default value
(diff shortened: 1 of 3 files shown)


src/pages/marketplace/generative/[id].tsx
```diff
@@ -317,6 +319,7 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
   return {
     props: {
       token: token,
+      urlQuery,
```
Comment (maerzhase) [inline src/pages/marketplace/generative/[id].tsx:322]: We should only return the query parameter we want to keep the seralised payload as small as possible

src/containers/Marketplace/GenerativeListings.tsx
```diff
@@ -45,15 +47,49 @@ function sortValueToSortVariable(val: string) {
 
 interface Props {
   token: GenerativeToken
+  urlQuery: Record<string, string>
```
maerzhase, line 50: We only need to provide the actualy initial sorting to the component

maerzhase (changes requested):
First of all: Thanks a lot for the support @Jobayer12 
There are a few things that need to be adjusted. 

Sorry for not being more specific in the issue description but there is already a context that stores user settings and also handles the whole localStorage part. We should add the setting `prefereredMarketplaceSorting` there. See here:
 https://github.com/fxhash/fxhash-website/blob/a52cfa60ad522b8413c16b9ca9f8ab30160aae28/src/context/Theme.tsx#L13-L31
 
 I attached a few more code comments.

src/pages/marketplace/generative/[id].tsx
Comment (Jobayer12) [reply src/pages/marketplace/generative/[id].tsx]: As per the direction of the comments, I made changes and push my code.

Questions:
1. I added "prefereredMarketplaceSorting" boolean field on ISettingsProperties interface. What to do next? Will it be functional later?

src/containers/Marketplace/GenerativeListings.tsx
Jobayer12: I made this change and please review my PR again.

src/pages/marketplace/generative/[id].tsx
Comment (maerzhase) [reply src/pages/marketplace/generative/[id].tsx]: Sorry but I cant provide a step by step guidance. you have to checkout how the setting context works and store the setting there.

src/containers/Marketplace/GenerativeListings.tsx
maerzhase: What I meant was not to serialize the query object but only the value what we need. we dont have any other query params for now so we can literally return the sorting value. 

rename sort -> sorting

maerzhase:
closing in favor of #575

PR #562, opened 2023-01-30
Hotfix/error on user profile navigation

maerzhase:
with the help of @louisholley superpower we figured it was an issue on the nextjs version :) 
further more fixed the warnings that `onClick` should be provided to children and not the `next/link` component
also updated the `@next/bundle-analyser` and the `next-eslint-config` to matching nextjs version


louisholley (commented):
lgtm! will QA when the preview is deployed

src/utils/hookts.ts
```diff
@@ -60,7 +60,7 @@ export const useClientEffect = (
   dependencies?: DependencyList
 ): void => {
   if (typeof window !== "undefined") {
-    useEffect(effect, dependencies)
+    useEffect(effect, dependencies) // eslint-disable-line
```
maerzhase, line 63: @GemN @louisholley this actually violates the hook rule that hooks can't be conditional. wdyt?

src/utils/hookts.ts
```diff
@@ -69,7 +69,7 @@ export const useClientAsyncEffect = (
   dependencies?: any[]
 ): void => {
   if (typeof window !== "undefined") {
-    useAsyncEffect(effect, dependencies)
+    useAsyncEffect(effect, dependencies) // eslint-disable-line
```
maerzhase, line 72: @GemN @louisholley this actually violates the hook rule that hooks can't be conditional. wdyt?

src/utils/hookts.ts
louisholley: i think it's fine in this case, `typeof window` will never go from undefined to undefined or vice versa so shouldn't give us any issues

src/utils/hookts.ts
GemN: yeah that's right, you can't use a condition before a hook. But a useEffect is already client side (per react documentation) so this part should be removed imo

louisholley (approved):

PR #568, opened 2023-02-09
fix: temporary styles fix for stylesheet order until nextjs update

GemN:


maerzhase (approved):
i clicked for a while and styles look good now again when navigating through the page. man what a beautiful hook :trollface:

PR #583, opened 2023-02-25
Feat/more query params sorting

louisholley:
adds `useQueryParamsSort` hook and uses it in each container with sorting options


maerzhase (approved):
@louisholley great job!
the only thing I stumbled upon now is the "Default marketplace sorting". Maybe we should eplicitly mention that its only the default for the listed artworks on the generative page and not for the global marketplace listings. 
Wdyt?

PR #584, opened 2023-02-27
Feat/articles collection sorting

louisholley:
adds "fake" sorting to collected articles as sort not currently supported in api


maerzhase (approved):
good job man

PR #617, opened 2023-03-20
fix: add toggable params label display, align feature label with content,  remove pointer-events: none and allow token interaction

GemN:


maerzhase (approved):
👍 🐰

PR #623, opened 2023-03-21
Feat/mint ticket alerts

louisholley:
adds generic user alerts (triggered once every 24 hour period using local storage) to implement this:

image

### DIFF
--- a/src/components/MessageCenter/Message.tsx
+++ b/src/components/MessageCenter/Message.tsx
@@ -10,10 +10,12 @@ interface Props {
 export function Message({ message, onRemove }: Props) {
   const [hidden, setHidden] = useState<boolean>(false)
 
+  const { type, title, content, keepAlive } = message
+
   const icon =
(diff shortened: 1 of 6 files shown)


GemN:
hi-ya some suggestions:

ux-wise:
- display the x by default (not on alert hover) so we understand it can be closed at a glance
- what do you think of the dark mode ? "see my tickets" can be hard to see. Maybe use purple in both case ?
Capture d’écran 2023-03-21 à 16 01 05

- `<br/>` on "see my tickets" to keep the action on one line
- bold on the number of tickets
- title change? Something describing the content inside, maybe something like "Mint tickets almost expired"
- IMO, I think the alert could be more precise - 48h before, 24h, 12h, 4h, 2h, 1h, 30min, 15min, 5min. 
It shouldn't be spammy that much

GemN (approved):
LGTM !

PR #628, opened 2023-03-22
Feature/attribute param creator

maerzhase:
After back and forth we decided to put the attribution in the meta table of the token. Anything more prominent wasn't appealing.

# Meta table attribution
For params token the col is called called "Params" with a tooltip. For other tokens we display "Minter"

Screenshot 2023-03-21 at 20 46 12
Screenshot 2023-03-21 at 20 46 08

# Discarded versions

Screenshot 2023-03-21 at 20 10 45
Screenshot 2023-03-21 at 20 10 41
Screenshot 2023-03-21 at 20 08 39
Screenshot 2023-03-21 at 20 08 32
image


louisholley (approved):
looks great man! nice one

PR #630, opened 2023-03-22
allow same hash on different params for variations

maerzhase:
We check if its params token we allow same hash with different params as a variant. If not params token hash must differ. Values are being displayed and truncated after 100chars

Screenshot 2023-03-22 at 11 43 23


louisholley (approved):
lgtm 👍

PR #640, opened 2023-03-23
allow user defined string maxlength

maerzhase:
this also requires a change in the snippet. So to test it we should merge file api PR to dev first.


louisholley (approved):
sweet!

maerzhase:
there is an issue with it. need to revisit

louisholley (approved):

PR #673, opened 2023-04-04
fix: collector opening multiple params view can use any ticket available from close to foreclosure

GemN:
passing a list of tickets to the mintv3 operation will take the first available and use it. 
if there are no tickets remaining, an precise error will be displayed


louisholley (approved):
looks good, will qa later

PR #709, opened 2023-04-17
add effect to temporary set html bg black

### LINKED ISSUE #674: set body bg black on mint with ticket view
image

to make the white header dark on mobile

maerzhase:
- fix #674

### DIFF
--- a/src/containers/MintWithTicket/MintWithTicketPage.tsx
+++ b/src/containers/MintWithTicket/MintWithTicketPage.tsx
@@ -274,6 +274,13 @@ export function MintWithTicketPageRoot({ token, ticketId, mode }: Props) {
     })
   }, [panelParamsRef, params, setHash, historyContext, withAutoUpdate])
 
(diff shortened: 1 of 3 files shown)


src/containers/MintWithTicket/MintWithTicketPage.tsx
```diff
@@ -274,6 +274,13 @@ export function MintWithTicketPageRoot({ token, ticketId, mode }: Props) {
     })
   }, [panelParamsRef, params, setHash, historyContext, withAutoUpdate])
 
+  useEffect(() => {
+    document.documentElement.classList.add(style.black)
```
louisholley, line 278: sweet man, maybe just add a comment to give this context in case we ever come across it in the future and wonder why it's there 😄

maerzhase: good call ;-)

PR #710, opened 2023-04-17
fix: random iterative cycler improved ux

GemN:
fix: random iterative cycler let image be clickable
fix: collab tokens in iterative cycler hide avatar as expected


maerzhase (approved):
@GemN looks good. i added `clearInterval` to effect cleanup

PR #717, opened 2023-04-19
fix: homepage events display all artists and event link

GemN:


louisholley (approved):

maerzhase:
@GemN i've added og + twitter meta

PR #736, opened 2023-04-28
Fix/live reveal bug

louisholley:


maerzhase (approved):
ty. just stumbled upon this error

PR #743, opened 2023-05-18
Feat/gentk contract from api

louisholley:
to avoid incorrect addresses being set in an issuer's royalties, we should defer to the API to inform us which gentk contract address is associated with a particular generative token


maerzhase (approved):
😎 👍

PR #748, opened 2023-05-25
Fix/accept offer gentk view

louisholley:
due to the missing offer.objkt relation in the main objkt query, offers weren't passing the typeguard in the shared accept offer code so the accept button wasn't being shown on offers for the individual gentk view. this fixes it and I believe won't fetch unnecessary data as the dataloader should cache results for the objkt itself in the api


maerzhase (approved):
sweet 1

PR #754, opened 2023-06-05
Feat/minting instructions meta

louisholley:
image

image

### DIFF
--- a/src/components/Collapsible/Collapsible.module.scss
+++ b/src/components/Collapsible/Collapsible.module.scss
@@ -1,0 +1,17 @@
+.container {
+  cursor: pointer;
+  display: flex;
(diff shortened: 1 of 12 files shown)


maerzhase (approved):
Thanks looks good! Only thing I am wondering have we decided whre this gets displayed? is it also on the fxhash website or only for live minting? if only for live-minting we might wanna add that to the description? if we want to display it in the website minting interface, we still need to add it there.

maerzhase (approved):
looks good @louisholley

maerzhase (approved):
@louisholley was wondering if we shouldn't validate the mint insturctions e.g. for max chars like description? apart from that looks good

PR #766, opened 2023-06-11
[WIP] Refacto param controls + support for sync params

ciphrd:
# Preamble

I was adding support for synced params, and I realized some architecture which bugged me in the way params and their controls was being handled. My main concern revolved around how the params definition object coming from the iframe context was being mutated and used as a state object for the controllers). I think it's always important to decouple the logic from the data, here the logic is the `params definition` and the data are the actual `param values`.

# Refacto

First of all, similarly to fx(lens), I created a [`useRuntime`](https://github.com/fxhash/fxhash-website/blob/d3815b0313818331f26fa3c14a87466b6d979760/src/hooks/useRuntime.ts) hook, which can be used as a source of state for all the iframe contexts running (it stores hash, minter, params, etc...). This ensures the any "token state" is isolated within the state provided by the hook, and provides a cohesive API throughout the app. 

This hook is now exposing, as distinct values, the `params definitions` and the `param values`, unmixed. A good chunk of this refacto was targeted at slightly updating the param controls to accommodate to this new logic. Now, param controls are **never** mutating the definition, and are using a state derived from (definition, values).

Then, `useRuntime` has been integrated into [`useReceiveTokenInfos`](https://github.com/fxhash/fxhash-website/blob/d3815b0313818331f26fa3c14a87466b6d979760/src/hooks/useReceiveTokenInfos.ts#L84), making sure this hook acts as a **connector** between the iframe and the state in the runtime hook. **This is still a work in progress, as some properties returned by the `useReceiveTokenInfos` object have still to get replaced so that runtime is used as the only source of truth through the application**. 

Support of "soft params sync" is now exposed by a function returned by `useReceiveTokenInfos`.

# TODO left
- [ ] runtime full state update at once (to prevent concurrency issues)
- [ ] fix the implementation at the mint > extra settings step
- [ ] fix the implementation in the params minting interface


ciphrd:
Improved upon the previous refaco, by adding a [`useRuntimeController`](https://github.com/fxhash/fxhash-website/blob/1b58368c873ff288da2b10f71a210bc308f2c828/src/hooks/useRuntimeController.ts) hook, to which **all control logic between the UI and the iframe is given.**.

This new hook receives some context details (project url, initial hash, minter, params, etc...), a pointer to an iframe and will handle any state management and logic related to the interactions between UI controls and the iframe, while exposing some utility functions and a state.

This architecture improves greatly on what we had previously:
- **separation of concern** between any UI and this logic
- **isolation** of the state in a controlled context, and all the logic related to how a frame should be synced based on interactions with the UI encapsulated at the same place; we don't have to scratch our head anymore to understand how data flows
- **clear data flow**: there used to be some weird side-effects & very complex data flow when manipulating params from the UI, it was okay for now but as we are going to iterate on fxhash it would have become a nightmare
- **consumption-agnostic** control-component, providing an un-opiniated interface so that all the consumer components can implement different UI flavors.

# Simple implementation diagram

Runtime Controller architecture

# Example 

This implementation demonstrate how clear the consumption of the `useRuntimeController` is: https://github.com/fxhash/fxhash-website/blob/1b58368c873ff288da2b10f71a210bc308f2c828/src/containers/MintGenerative/StepCheckFiles.tsx

louisholley (approved):

PR #767, opened 2023-06-12
hide references to price/wallet balance for free live minting

louisholley:


maerzhase (approved):
jeah cool noticed this last week too

PR #772, opened 2023-06-13
Fixed sandbox not working with the new runtime Controller

ciphrd:


maerzhase (approved):
the sandbox works now. the looping stuff was fixed but sometimes we still get the "failed to fetch" error on the worker.

Screenshot_2023-06-14_at_16 44 49

PR #774, opened 2023-06-14
Fix/runtime state update loop

maerzhase:
should fix most of the quirky and "race conditions" we experienced (especially on old snippet projects) 

- this view should work on this branch without looping: https://dev.fxhash-dev.xyz/generative/slug/test-for-tickets/explore-params?fxhash=null&fxparams=


louisholley (approved):
huuuge

PR #780, opened 2023-06-16
hide winter for v1 marketplace contract

louisholley:
winter was never implemented for the v1 marketplace - may as well hide the option for v1 listings!


maerzhase (approved):
sick!

PR #790, opened 2023-06-19
update minter in state as passed props change

louisholley:


maerzhase (approved):
seems better than before :>

PR #794, opened 2023-06-19
Feat/pass minter to params view

louisholley:


louisholley:
closing because i think we need better UX to include this change

PR #802, opened 2023-06-23
iteration fixes // params context switch

louisholley:


src/containers/MintWithTicket/Panel/PanelContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Switch } from "components/Input/Switch"
+
+export interface PanelContextProps {
+  context: "minting" | "standalone"
+  onChangeContext: (c: "minting" | "standalone") => void
+}
+
+export function PanelContext({ context, onChangeContext }: PanelContextProps) {
+  return (
+    <PanelGroup title="Context" description="Useful description...">
```
louisholley, line 12: any suggestions for description here? @maerzhase

maerzhase: i'd suggest something like: "Toggle the execution context of the artwork between "minting" and "standalone""

PR #807, opened 2023-06-29
fix use ticket with reserves bug

louisholley:
seems like when the user has reserves for a project and click use ticket, we are attempting to show the reserve dropdown (which is hidden because you're using a ticket) so nothing happens in the UI...this should fix


maerzhase (approved):
good catch friend!

PR #817, opened 2023-07-04
Fix/old snippet params variations

louisholley:
- for e.g. https://www.fxhash.xyz/generative/26248, the variations are broken as we are not correctly passing the params definition to be serialized, so it's opening with random params and changing on each refresh. this PR should fix that issue

- also fixes an issue with the old snippet where params weren't serializing correctly as we were passing a default version so the incorrect serialization code was being used

### DIFF
--- a/src/components/GenerativeToken/GenerativeArtwork.tsx
+++ b/src/components/GenerativeToken/GenerativeArtwork.tsx
@@ -14,8 +14,8 @@ import { SettingsContext } from "../../context/Theme"
 import { ipfsGatewayUrl } from "../../services/Ipfs"
 import { Image } from "../Image"
 import { useReceiveTokenInfos } from "hooks/useReceiveTokenInfos"
-import { getGenerativeTokenUrl } from "utils/generative-token"
 import { ButtonExploreParams } from "components/Button/ButtonExploreParams"
+
 interface Props {
   token: Pick<
     GenerativeToken,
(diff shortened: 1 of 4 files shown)


src/hooks/useRuntime.ts
```diff
@@ -137,7 +137,7 @@ export function useRuntime(initial?: Parameters): IRuntimeContext {
       params:
         definition.params?.map((p: FxParamDefinition<FxParamType>) => ({
           ...p,
-          version: definition.version || "0",
```
louisholley, line 140: if we pass a default here, we will not be using the correct code to serialize params for the old snippet

maerzhase: !!!!! you are the man that you identified this issue. jeah we really shouldn't set a "0" for default hahahaahh

maerzhase (approved):
man what a hell of a debugging session!!! really nice that we found it in the end phew...

src/components/GenerativeToken/GenerativeArtwork.tsx
```diff
@@ -49,7 +49,10 @@ export function GenerativeArtwork({
   const settings = useContext(SettingsContext)
   const artworkIframeRef = useRef<ArtworkIframeRef>(null)
 
-  const { params, onIframeLoaded } = useReceiveTokenInfos(artworkIframeRef)
+  const { paramsDefinition: paramsDefinitionFromIframe, onIframeLoaded } =
+    useReceiveTokenInfos(artworkIframeRef)
+  const paramsDefinition =
+    paramsDefinitionFromIframe || token.metadata.params.definition
```
louisholley, line 55:
```suggestion
    paramsDefinitionFromIframe || token.metadata.params?.definition
```

PR #818, opened 2023-07-05
temporarily filter inactive redeemables

louisholley:
we can't delete redeemables in the backend rn so to avoid having multiple broken redeemables displayed in the frontend, this pr filters out all redeemables that have been made inactive from the UI. once we implement proper deletion of redeemables these manual fixes can be removed


maerzhase (approved):
👍

PR #820, opened 2023-07-05
Feat/retry fetch random seed

louisholley:
retry fetching the random seed after minting up to 5 times before gracefully handling the error

### DIFF
--- a/src/hooks/useFetchRandomSeed.ts
+++ b/src/hooks/useFetchRandomSeed.ts
@@ -4,7 +4,10 @@ import { useEffect, useState } from "react"
 
 const MAX_RETRIES = 5
 
-export function useFetchRandomSeed(opHash: string | null) {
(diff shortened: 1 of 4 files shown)


src/hooks/useFetchRandomSeed.ts
```diff
@@ -1,15 +1,32 @@
...
+      if (retries < MAX_RETRIES) {
+        setRetries(retries + 1)
+      } else {
+        setError(new Error("Maximum retry attempts reached"))
+      }
+    },
   })
 
+  useEffect(() => {
+    if (data === null && retries < 3) refetch()
```
maerzhase, line 27: you probably want to use `MAX_RETRIES` here?

maerzhase (approved):
only this one minor thing. other than that should be good

PR #822, opened 2023-07-05
Enhancement/context selection as buttons below artwork

maerzhase:
The context selection was moved below the artwork as buttons. This should solve some confusion for artists that were under the impression that the context selection would somehow be related to the settings of the preview. 

Screenshot 2023-07-05 at 19 40 29

# How to test

- Open the preview (see comment below from vercel bot)
- Navigate to the "Mint generative Token" flow
- Upload an artwork
=> The context buttons show now be displayed below the artwork

Project to upload: [project.zip](https://github.com/fxhash/fxhash-website/files/11965999/project.zip)

# Implementation details

- I added an `asButtons` property to the `ContextTest` component that renders the context selection as buttons instead of a select menu
- I fixed an issue that we were actually not rendering the artwork on small breakpoints anymore 🤣

### DIFF
--- a/src/components/Testing/ContextTest.module.scss
+++ b/src/components/Testing/ContextTest.module.scss
@@ -2,6 +2,15 @@
   small {
     font-weight: bold;
   }
+  .buttons {
+    margin-top: var(--spacing);
+    display: flex;
+    justify-content: space-between;
+    gap: var(--spacing);
+  }
+  .button {
+    flex-grow: 1;
+  }
   .select {
     width: 100%;
   }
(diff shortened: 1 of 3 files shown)


IsabelleKloeck (approved):
Hi Markus, tested. Everything working as expected.

louisholley (approved):
lgtm! should we do something about the context toggle in the minting view here as well (some people have complained that there isn't enough explanation for what the context switch does there) or will that be tackled as another task?

maerzhase:
@louisholley i think we should make a separte issue for it. i will add it to the requests + ideas

leafxhash:
For the mobile view, would you add some additional space below the button? Except that, everything looks good to me. Thanks. 
Screenshot 2023-07-06 at 2 51 27 PM

PR #823, opened 2023-07-06
Enhacement/event section redesign

maerzhase:
This PR implements the "event section redesign".
- Reorder sections on the homepage
- Redesign event cards

# UI/UX details

There are a few logical operations happening to display the information on the card:

#### Online vs. Offline Event Location

In the designs we see that an event can be online and somewhere located. 

Screenshot 2023-07-06 at 15 04 53

As of now we don't store "multiple locations" and/or offline vs. online event informations. Therefore we are either displaying the location of the event if it has one. If the event has no location specified it will automatically display as "online".

#### Event Time 

In the designs we are showing the different event dates in two ways. Either as a range from start to end OR the start date with the start time. 

Screenshot 2023-07-06 at 15 07 27

I implemented the following logic: 
- When event is ongoing we display the start date and end date
- When event is upcoming we display only the start date with the time it starts

Does that make sense?

# How to test
- Open the preview (see comment below from vercel bot)
- Navigate to the homepage
  - Confirm the order of all sections
- Checkout the Event section
  - Confirm new design
  - Confirm UI/UX details (see above)
  - Confirm layout on multiple breakpoints/screensizes/devices

# Implementation details

- I refactored the documentation section on the homepage into its own component for more flexibility in ordering the components
- Removed a lot of unused code from the old `CardEvent` component, because the new version is much more minimal

### DIFF
--- a/src/components/Card/CardEvent.module.scss
+++ b/src/components/Card/CardEvent.module.scss
@@ -9,7 +9,6 @@
   background: var(--color-gray-vvvvlight);
   color: var(--color-black);
   text-decoration: none !important;
-  height: 350px;
   .cardHeader {
     aspect-ratio: 293 / 286;
     height: 150px;
@@ -29,6 +28,7 @@
   font-size: 16px;
   font-weight: 700;
   line-height: 20px;
+  margin-bottom: var(--spacing-xl);
 }
 .date, .location {
   font-size: 14px;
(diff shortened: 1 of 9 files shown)


leafxhash (approved):
Tested, it looks so perfect. Thanks, Markus.

IsabelleKloeck (approved):
Looks great! Thank you!

louisholley (approved):
lovely stuff

PR #826, opened 2023-07-14
Add tracking via matomo and plausible

maerzhase:
This PR adds two tracking services:
- Matomo
- plausible

> We are integrating two services for a test-period to evaluate which service serves our needs the best.

Each tracking service is initialised within their hook. The hooks exposes a streamlined API that is an abstraction on the actual services's API. For now we are tracking page views only. 

# Implementation details

- A new environment variable is expected: `process.env.NEXT_PUBLIC_DEPLOY_ENV`
  - Expected values are `production | staging | dev | local`
  - Tracking services will only be initialised on `NEXT_PUBLIC_DEPLOY_ENV === "production"`
  - When not on production page-views are logged to the console
- For each service a hook was added:
  - `useMatomo` initialises the matomo service and exposes the `trackPageView` method
  - `usePlausible` initialises the plausible service and exposes the `trackPageView` method 
- A hook `useTrackPageView` was added that receives a handler that is called on mount and on `router.events.on("routeChangeComplete")`
- The tracking services are initialised and hooked up in `src/containers/Root.tsx`

# TODO

- [ ] @louisholley we need to set the `NEXT_PUBLIC_DEPLOY_ENV` on vercel for the different environments


src/hooks/useMatomo.ts
```diff
@@ -0,0 +1,30 @@
+import Matomo from "matomo-tracker"
+import { useMemo, useCallback } from "react"
+import { TrackingService } from "./usePlausible"
+
+let _matomo: any
+
+const DEBUG = process.env.NEXT_PUBLIC_DEPLOY_ENV !== "production"
```
maerzhase, line 7: We are using the new `NEXT_PUBLIC_DEPLOY_ENV` variable to determine if the service should be live or not.

louisholley (approved):
lovely! i don't have access to vercel so we'll need @ciphrd to add the env var

maerzhase:
will be integrated through the monorepo

PR #829, opened 2023-07-16
set allow-modals and * allow_list for allowed features

maerzhase:
It was reported that camera wasn't working. There were two issues:

- We need `allow-modals` so the iframe can actually trigger the modal to set permission for camera usage
- We also needed to set the allow_list for the features we allow for cross origin to allow in top-level browsing contexts and in nested contexts (iframes)

Reference: https://blog.addpipe.com/camera-and-microphone-access-in-cross-oirigin-iframes-with-feature-policy/

Aside from that we are now also always just rendering one iframe per breakpoint instead of hiding them via css


ciphrd (approved):
LGTM!

PR #833, opened 2023-07-18
Feature: inputbytes as url hash when version lower than 3.2.0

maerzhase:
This PR address the issue that long inputbytes don't work on certain CDN services because it can reach maximum url length. therefore we use the hash for inputbytes in order to not send the long url to the server
- stores the `snippetVersion` in the token metadata
- constructs the url based on the snippet version
  - <= v3.2.0 puts the fxparams into query params `?fxparams=....`
  - \>v3.2.0 puts the fxparams into the hash `#....`


ciphrd:
✅ LGTM

Closing in favor of #834 which bundles all the changes related to the snippet 3.3.0

PR #839, opened 2023-07-27
add redeemable phone number input

louisholley:


pradel (approved):
LGTM, but this pr should be moved to the monorepo as the production website is now deployed from it :)

